PR #1, opened 2015-05-20
Added math.poly and math.linear functionality to support mcad_analysis

anna-dodd:
Introduced angle and normSq functions to Vector and VectorN classes, introduced velocity and acceleration functions to OneD, and created a new constructor for TwoD to initialize coefficients with a std::vector<OneD<_T> >. These additions allow mcad_analysis files to now build with CODA instead of older source files that included these missing functionalities. 

Trac ticket #3

### DIFF
--- a/modules/c++/math.linear/include/math/linear/Matrix2D.h
+++ b/modules/c++/math.linear/include/math/linear/Matrix2D.h
@@ -969,11 +969,8 @@ public:
         }
         return perm;
     }
-    /*
-     * Find the square of the L2 norm
-     * Sum of squares of the vector elements
-     */
-     _T normSq() const
+    // Addition made 5-13-15:
+    _T normSq() const
     {
         size_t sz = mM * mN;
         _T acc(0);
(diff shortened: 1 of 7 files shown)


asylvest:
Everything looks correct.  Some review comments...

Matrix2D.h
==========
- For the return value of normSq(), you don't need the cast
- For the return value of norm(), use static_cast<_T> instead

MatrixMxN.h
===========
- For the return value of normSq(), you don't need the cast

Vector.h
========
- For the angle() implementation, collapse lines 239-241 to one line using std::max() and std::min().  Once you do that, 'val' can be const.  Don't put ()'s around the return value.

VectorN.h
=========
- Ditto the angle() comments above.  Additionally, pass Like_T in by const ref.  And change comment to say "this vector and v"
- If using stuff in std:: like std::acos, need to include <cmath> rather than <math.h>.  Ditto for Vector.h?

OneD.hpp
========
- For velocity() and acceleration, you shouldn't need the (*this).  Also I don't think you need the () at all - can't you just say 'return derivative()(x)'?

TwoD.h
======
- New constructor: Pass that vector in by const ref (we don't want to make a copy - could be expensive).  Also, put those {} each on their own line.  And add a newline before and after the constructor.

anna-dodd:
I made changes to the files according to the review comments above.

asylvest:
Address the review comments above.  Also, I can't automatically merge in this pull request which means some other stuff must have happened on the master... can you merge the master into your branch such that I can just automatically merge this pull request in?

asylvest:
Everything is good except the merge conflicts... can you get this branch synced up with master so that I can do a one click merge?  I think that will involve you trying to do a merge of the master into your branch (** so you want to do a checkout of your branch and then run the git merge command, pointing it to the master**).  I'm guessing when you try to run that merge command you will get conflicts that you'll need to resolve.

PR #9, opened 2015-05-26
Soname crossplatform

chvink:
Fixed issue when configured with --shared and cleaned up process_swig_linkage in general


asylvest:
Yeah this is what I was worried about... may be easier to make a new branch and just merge your last two commits into it and then do a pull request for that (this pull request can't be automatically merged).

PR #12, opened 2015-05-27
Tests for logger python bindings

asylvest:


asylvest:
Looks good - nice work tracking down how to get this working.  Just one more thing... can you include FileHandler in the Swig interface file and then add to the test script showing that this works properly?  This *should* just work out of the box without any other craziness I think.

PR #13, opened 2015-05-28
Added tests for the new math.linear functions

anna-dodd:


modules/c++/math.linear/tests/test_angle.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include <TestCase.h>
+#include <import/math/linear.h>
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+// Explicitly recreating the Vector3D implementation of angle function
+double vec3DAngle(const math::linear::VectorN<3>& v1, const math::linear::VectorN<3>& v2)
```
asylvest: Suppose on one platform there was a bug in our actual implementation of the angle calculation (we divided by 0 or did something else weird and unsafe).  We wouldn't catch it with this approach since it would do the same wrong thing in this function and in the actual Vector implementation.  So rather than this, I'd pick 5-10 known cases and add these with explicit hard-coded expected results.  You could do that by just running your current implementation and printing out the current values and then using those as the hard-coded values to use.

modules/c++/math.linear/tests/test_norm_v.cpp
```diff
@@ -0,0 +1,127 @@
...
+    TEST_ASSERT_ALMOST_EQ(expectedNorm3, actualNorm3);
+
+    double expectedNorm4 = std::sqrt(expectedNormSq4);
+    double actualNorm4 = v4.norm();
+    TEST_ASSERT_ALMOST_EQ(expectedNorm4, actualNorm4);
+}
+double manualNormSq(const math::linear::Vector<double>& v)
+{
+    return (v[0] * v[0]) + (v[1] * v[1]) + (v[2] * v[2]);
+}
```
asylvest: Move this to an unnamed namespace at the top of the file

modules/c++/math.linear/tests/test_norm_vN.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    double expectedNorm3 = std::sqrt(expectedNormSq3);
+    double actualNorm3 = v3.norm();
+    TEST_ASSERT_ALMOST_EQ(expectedNorm3, actualNorm3);
+
+    double expectedNorm4 = std::sqrt(expectedNormSq4);
+    double actualNorm4 = v4.norm();
+    TEST_ASSERT_ALMOST_EQ(expectedNorm4, actualNorm4);
+}
+double manualNormSq(const math::linear::VectorN<3> v)
```
asylvest: Move this to an unnamed namespace at the top of the file

modules/c++/math.linear/tests/test_norm_v.cpp
```diff
@@ -0,0 +1,127 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <TestCase.h>
+#include <math/linear/Vector.h>
+#include <math.h>
+
```
asylvest: Put everything except main() in unnamed namespace

modules/c++/math.linear/tests/test_norm_vN.cpp
```diff
@@ -0,0 +1,129 @@
...
+ *
+ */
+
+#include <TestCase.h>
+#include <import/math/linear.h>
+#include <math.h>
+
+
+//Vector3D norm()--> normSq()
+TEST_CASE(VaryingDimension)
```
asylvest: Put everything except main() in unnamed namespace

modules/c++/math.linear/tests/test_norm_vN.cpp
```diff
@@ -0,0 +1,129 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <TestCase.h>
+#include <import/math/linear.h>
```
asylvest: Use actual headers you need rather than the import header

modules/c++/math.linear/tests/test_angle.cpp
```diff
@@ -0,0 +1,92 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <TestCase.h>
+#include <import/math/linear.h>
```
asylvest: Use actual headers you need rather than the import header

anna-dodd:
I added in the unnamed namespaces and added in those hard coded tests to test_angle

PR #14, opened 2015-06-02
Add opencv

adam-beauchamp:


jmrandol:
Waiting on security for OpenCV approval.  There seems to be an issue with the latest version. @adam-beauchamp have you been keeping up on this?

jmrandol:
This hasn't been updated in a while, is it still relevant?  If so, it would need to be updated to be built with CMake

adam-beauchamp:
All of the work in this PR was to duplicate cmake's operations in *waf* 🤢 ; I'm not convinced that having a coda driver for this library is worthwhile anyway: every driver ends up bloating coda's interface to include all of the configuration options for each of the component libraries. OpenCV is a massive library with *tons* of build options which would bloat coda's configuration interface immensely.

Terminating with extreme prejudice.

PR #17, opened 2015-06-04
Cpp11 switch

jakedebolt:
This allows c++11 usage to be determined at configure-time. There is also support for allowing wscripts to force c++11 usage via the 'force_cpp11support' parameter.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -686,6 +686,8 @@ def options(opt):
                    default=False, help='Treat compiler warnings as errors')
     opt.add_option('--enable-debugging', action='store_true', dest='debugging',
                    help='Enable debugging')
+    opt.add_option('--enable-cpp11', action='store_true', default=False, dest='enablecpp11',
+                   help='Enable C++11 features')
     #TODO - get rid of enable64 - it's useless now
     opt.add_option('--enable-64bit', action='store_true', dest='enable64',
                    help='Enable 64bit builds')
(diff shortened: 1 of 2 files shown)


build/build.py
```diff
@@ -813,7 +815,9 @@ def configureCompilerOptions(self):
             config['cxx']['optz_fastest']   = '-O3'
 
             gxxCompileFlags='-fPIC'
-            if cxxCompiler == 'g++' and gccHasCpp11():
+            botherChecking = self.env['cpp11support'] or self.env['force_cpp11support']
```
asylvest: I don't think you need `self.env['force_cpp11support']`.  If an individual project wants to force C++11 support, they can do so in their top-level wscript by putting `self.env['cpp11support']` in their configure section.  I don't think you need two separate environment variables.

build/build.py
```diff
@@ -813,7 +815,9 @@ def configureCompilerOptions(self):
...
+            if cxxCompiler == 'g++' and botherChecking and gccHasCpp11:
```
asylvest: Don't you need the (): `gccHasCpp11()`?

build/build.py
```diff
@@ -813,7 +815,9 @@ def configureCompilerOptions(self):
...
+                print 'why'
```
asylvest: Don't need this

modules/c++/sys/wscript
```diff
@@ -74,16 +67,7 @@ def configure(conf):
...
-	t2 = Utils.str_to_dict(output11 or '')
-	for k, v in t2.iteritems():
-		v = v.strip()
-		if v == 'True':
-			v=True
-		elif v == 'False': 
-			v=False;
-		conf.define(k.upper(), v)
-        
+        conf.check_cxx(fragment = cpp11_str, execute=1, msg='Checking for C++11 support', define_name='__CPP11', mandatory=False)
```
asylvest: I must have missed this part the first time... let's make the define name something like CODA_CPP11.  We want to make sure we don't conflict with some reserved preprocessor thing.

PR #18, opened 2015-06-05
bytesToWrite now is calculated against a size_t.

clydestanfield:
Same issue as reader, but with the writer. The bytesToWrite needs to be calculated using a size_t instead of a DWORD.

### DIFF
--- a/modules/c++/sys/source/FileWin32.cpp
+++ b/modules/c++/sys/source/FileWin32.cpp
@@ -91,15 +91,15 @@ void sys::File::readInto(char *buffer, Size_T size)
 
 void sys::File::writeFrom(const char *buffer, Size_T size)
 {
-    static const DWORD MAX_WRITE_SIZE = std::numeric_limits<DWORD>::max();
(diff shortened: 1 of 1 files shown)


modules/c++/sys/source/FileWin32.cpp
```diff
@@ -91,15 +91,15 @@ void sys::File::readInto(char *buffer, Size_T size)
...
     size_t bytesRemaining = size;
     size_t bytesWritten = 0;
 
     while (bytesWritten < size)
     {
         // Determine how many bytes to write
-        const DWORD bytesToWrite =
-            std::min<DWORD>(MAX_WRITE_SIZE, bytesRemaining);
+        const DWORD bytesToWrite = static_cast<DWORD>(
+            std::min<size_t>(MAX_WRITE_SIZE, bytesRemaining));
```
asylvest: Can remove this <size_t> right?

clydestanfield:
The <size_t> has been removed. I had it in the reader as well, so I removed it from both locations.

asylvest:
All set

PR #20, opened 2015-06-10
Complex benchmark

dj-snyder:


modules/c++/math/tests/complexBenchmark.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <cstdlib>
+#include <ctime>
+#include <iostream>
+#include <fstream>
+#include <iomanip>
+#include <sstream>
+#include <complex>
+#include <sys/StopWatch.h>
+
+std::complex<float> getMeanWComplex(sys::RealTimeStopWatch& wtch, std::complex<float>* in, long long sze, double& duration)
```
asylvest: Always use size_t rather than long long for this kind of thing... size_t will always be big enough to hold an array in memory

modules/c++/math/tests/complexBenchmark.cpp
```diff
@@ -0,0 +1,159 @@
...
+    meanQ /= sze;
+    
+    duration = wtch.stop();
+    std::complex<float> tmp(meanI, meanQ);
+
+    return tmp;
+
+}
+
+void print(std::ostream& out, long long sze, std::complex<float> meanOne, std::complex<float> meanTwo, double durOne, double durTwo)
```
asylvest: I am going to buy you a tshirt with "80 CHARACTERS" printed on it :o)

modules/c++/math/tests/complexBenchmark.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+int main(int argc, char** argv)
+{
+    if (argc != 4)
+    {
+        std::cerr << "ERROR, incorrect calling" << std::endl;
+        std::cerr << "use ./test <InitialArraySize>  <growthFactor> <numberOfIterations>" <<std::endl;
+        return 1;
+    }
+    long long sze = atoi(argv[1]);
```
asylvest: Use str::toType() rather than atoi() for stuff like this

modules/c++/math/tests/complexBenchmark.cpp
```diff
@@ -0,0 +1,159 @@
...
+    float Real =  rand() % 100 + 1;
+    float Imag =  rand() % 100 + 1;
+
+    //std::cout << "Assigning first spot" << std::endl;
+    arr[0] = std::complex<float>(Real, Imag);
+    long long j = 1;    
+    for (long long i = 0; i < numIter; ++i)
+    {
+
+        //std::cout << "creating array" << std::endl;
```
asylvest: Remove the couts and old comments throughout

modules/c++/math/tests/complexBenchmark.cpp
```diff
@@ -0,0 +1,159 @@
...
+    out.setf(std::ios::fixed);
+    out.setf(std::ios::left);
+    
+    outFile << std::setprecision(5);
+    outFile.setf(std::ios::fixed);
+    outFile.setf(std::ios::left);
+
+    //Initialize the array ahead of time and then fill it incrementally
+    const unsigned long long MAX_SIZE = 10E10 / (sizeof( std::complex<float>));
+    std::complex<float>* arr = new std::complex<float> [MAX_SIZE];
```
asylvest: Use a std::vector or a mem::ScopedArray

modules/c++/math/tests/complexBenchmark2.cpp
```diff
@@ -0,0 +1,160 @@
+/* Users guide
```
asylvest: Can't these two benchmarks be combined into one where you just optionally loop N times?

modules/c++/math/wscript
```diff
@@ -2,6 +2,7 @@ NAME            = 'math'
 MAINTAINER      = 'asylvest@users.sourceforge.net'
 VERSION         = '0.1'
 USELIB          = 'MATH'
+MODULE_DEPS     = 'sys'
```
asylvest: Change this to TEST_DEPS = 'sys'.  This will make only the test depend on sys but not the module itself.

dj-snyder:
Ready for code Review, changes were made

asylvest:
`str::toType<size_t>(argv[1])` compiles for me... where did this not compile for you?

dj-snyder:
I was forgetting the <size_t>, will change

modules/c++/math/include/math/Constants.h
```diff
@@ -23,6 +23,31 @@
 #ifndef __MATH_CONSTANTS_H__
 #define __MATH_CONSTANTS_H__
 
+#include <sys/sys_config.h>
+
+#ifdef __CODA_CPP11
+#include <math.h>
+namespace math
+{
+struct Constants
```
asylvest: We don't want to define all the constants twice... can you figure out some way to avoid this?
Also, I don't think you need the word static with constexpr do you?

dj-snyder: I tried not defining them here, but because of the way that constexpr works, it requires you to give it a value as soon as you use it.  It basically attempts to evaluate a constexpr variable at compile time.  You also do still need static or else you can't access them without instantiating the class I believe.

asylvest: Hmm... yeah I guess that's true.  So then there's no real solution aside from `#define` which defeats the purpose of trying to have static constants.  Bah.

Constants.cpp should not redefine the constants for the C++11 case though right (these are unrelated constants as far as the compiler is concerned - they're defined outside of the Constants class).

dj-snyder: I'm not quite sure what/ if you are asking, but I removed the double definition whenever __CODA_CPP11 is defined.

asylvest:
Yep you got it

PR #23, opened 2015-06-10
Added Line2D

anna-dodd:
This is not completely ready to be merged obviously, there is one area with the implementation of a Polygon2D function I want to address, in bool pointInPolygon(const Line2D::Point& P), where a comparison between two doubles is necessary and I am unsure what our margin of error should be (epsilon or something larger?). See my comments on Polygon2D.h. I added in tests with some hard-coded examples but I could add more.


anna-dodd:
I updated Line2D's intersection implementation for better functionality (saving some try/except blocks) for Polygon2D's intersection function

modules/c++/math.linear/source/Line2D.cpp
```diff
@@ -156,7 +154,18 @@ Line2D::Point Line2D::intersection(const Line2D& rhs) const
...
+    return true;
+}
+Line2D::Point Line2D::intersection(const Line2D& rhs) const
+{
+    Point P(0,0);
+    if (!intersection(rhs, P))
+    {
+        throw except::Exception(Ctxt("Lines do not intersect"));
+    }
+    else {
```
asylvest: This brace should be on a newline

asylvest:
Just one trivial change with that brace and then Line2D is all set.

But... this branch still has the Polygon2D updates in it right?  So, need to remove those files and revert the wscript change so that all that we would merge in is the Line2D stuff.

anna-dodd:
I took care of changing that formatting and removed Polygon2D stuff

PR #32, opened 2015-07-10
normL1() to area()

anna-dodd:


asylvest:
Had waited too long on this and things got out of sync... redid this under #114

PR #33, opened 2015-07-11
Python install update

asylvest:


modules/python/wscript
```diff
@@ -1,3 +1,5 @@
+import os
```
asylvest: Is this change really needed?

asylvest:
@chvink 
@chvink 
- test_math_poly.py needs to be updated
- Does not build on Windows (use something built-in in Python to create the empty file rather than the Unix 'touch' command)
- So if I follow the current logic, you need the try/except because every Swig target is going to try to make ```__init.py___``` so you'll get some failures?  Rather than doing it this way, I think you should be able to make the target and then use the targets_to_add feature to say this current Swig target should also eventually build the ```__init.py___``` target.  This way waf will see a lot of Swig targets depend on this but only try to actually execute it once.

asylvest:
Also, make sure to merge master into this branch and verify all is still well (I just made a couple updates related to the Python bindings and merged them into master... they shouldn't conflict with your stuff as long as git doesn't do something dumb).

chvink:
The last few updates should take care of these issues.

PR #42, opened 2015-07-20
Improve the .gitignore

porglezomp:
Ignore untarred driver folders
Ignore some extra files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,24 @@
 *~
 *.pyc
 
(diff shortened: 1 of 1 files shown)


chvink:
Might as well add .swp while you're at it.

porglezomp:
It had `*.swp` already but the diff doesn't show enough context for it to be visible. If you expand the collapsed line at the top you can see it.

PR #46, opened 2015-08-17
Types row col sqrt fix for windows

mbojrab:
Visual Studio 2010 was unable to resolve the sqrt for size_t or sys::SSize_T, so because this isn't heavily used and types doesn't depend on sys, we perform a cast to double to make the template we intend to use explicit. This is a little unfortunate.


asylvest:
I made one more commit to make the comments more explicit and there was one static_cast that could be cut out too.

PR #48, opened 2015-08-21
Added ability to rename Swig'd smart pointer in macro

chvink:
This is needed to prevent name conflicts across modules.


asylvest:
Can you check in the generated files too (python/mem/source/generated*)?

chvink:
I'm not seeing any difference in them, the macros probably don't get to the generated files.

asylvest:
Oh duh - right.

PR #51, opened 2015-08-21
Adding a typemap for Off_T

chvink:
We need this for our bindings elsewhere. There are several sets of typemaps here because I suspect the one set would have trouble on windows. I'm not sure why but sys::Off_T does not actually match usages of sys::Off_T. 

Adding the typemaps didn't seem to change the cxx and py wrappers either.


asylvest:
If you say so

PR #57, opened 2015-10-05
Sys exec terminate

mbojrab:
Update to provide access to the subprocess handle. In the regular popen() call this handle is closed, so  if the parent dies it has no access to close the child as well. The pclose additionally calls wait on the subprocess which blocks the host until the child has completed. 

We need a way to instantly kill the subprocess on the event where the parent dies early and need to cleanup. The only way to perform this is to roll your own popen/pclose which do not do the faus pas above, and instead provide all the functionality exactly as we need it.

On Windows I made the executive decision (after conversing with Jeff) to only provide access to "r" mode which is generally what you will need. The "w" would allow writing to the stdin from file or another means, but decided instead to simplify this and connect the parent's stdin to this pipe, so its essentially a pass-thru. StdOutput is connected as expected, and is working.

I tested this on three test cases:

"PipeStreamTest LongRunningJob" -- which runs forever and requires a user initiated cancel (ctrl+c)

"PipeStreamTest python printHello.py" -- prints "HelloWorld" and closes

"PipeStreamTest python" -- runs the python console to the subprocess, and tests all stdin/stdout to the console. Ctrl+c and quit() cancels the process equally.


modules/c++/sys/source/ExecUnix.cpp
```diff
@@ -24,9 +24,115 @@
 #if !defined(WIN32)
 
 #include <sys/wait.h>
-#include "sys/Exec.h"
+#include <str/Manip.h>
+#include <sys/Exec.h>
 
-int sys::ExecPipe::closePipe()
+#define READ_PIPE  0
```
asylvest: BOO for `#define`s.  Make these static consts in an unnamed namespace.

modules/c++/sys/source/ExecUnix.cpp
```diff
@@ -24,9 +24,115 @@
...
+{
+
+FILE* ExecPipe::openPipe(const std::string& command,
+                         const std::string& type)
+{
+    register FILE* ioFile = NULL;
+    int pIO[2];
+
+    //! create the IO pipes for stdin/out
+    if (pipe(pIO) < 0) { return NULL; }
```
asylvest: Curly braces and body go on their own lines

modules/c++/sys/source/ExecUnix.cpp
```diff
@@ -24,9 +24,115 @@
...
+
+    //! fork a subprocess for running our command --
+    //  here we use the user-defined pid, which is one major
+    //  differences between this and the normal popen()
+    switch (mProcess = fork())
```
asylvest: For clarity, avoid assignments inside other statements.  So assign `mProcess`, then `switch` on it.

modules/c++/sys/source/ExecUnix.cpp
```diff
@@ -24,9 +24,115 @@
...
+    {
+        case -1:
+        {
```
asylvest: Curly braces not needed inside this case statement

modules/c++/sys/source/ExecWin32.cpp
```diff
@@ -23,10 +23,97 @@
 
 #if defined(WIN32)
 
-#include "sys/Err.h"
-#include "sys/Exec.h"
+#include <sys/Exec.h>
+#include <str/Manip.h>
 
-int sys::ExecPipe::closePipe()
+#define READ_PIPE  0
```
asylvest: Avoid `#define`s here and reevaluate with `pclose()`, etc. below

modules/c++/sys/source/ExecWin32.cpp
```diff
@@ -23,10 +23,97 @@
...
+#define pclose _pclose
+#define fileno _fileno
+
+namespace sys
+{
+
+FILE* ExecPipe::openPipe(const std::string& command,
+                         const std::string& type)
+{
+    register FILE* ioFile;
```
asylvest: Why `register` keyword (both here and in Unix version)?  Don't think this is needed...

asylvest:
This functionality is always messy to code up (which is why it's good we'll have it in one spot in CODA :o) ).  I am assuming the implementation itself has been validated (sounds like it) so I'm just verifying style.  Just a couple nits.

modules/c++/sys/source/ExecUnix.cpp
mbojrab: Can you clarify what you mean here? Are you saying this because it returns?

modules/c++/sys/source/ExecWin32.cpp
mbojrab: the "register" keyword was used in the real popen source implementation, so I added this here. The application work regardless of its presence, so I will remove it now.

modules/c++/sys/source/ExecUnix.cpp
asylvest: You only need to put curly braces below a case statement if you need to scope something (if you declare a new variable).  In this case you're not declaring any variables in the case section so the braces aren't needed.

modules/c++/sys/source/ExecUnix.cpp
```diff
@@ -24,9 +24,120 @@
...
 
 #include <sys/wait.h>
-#include "sys/Exec.h"
+#include <str/Manip.h>
+#include <sys/Exec.h>
 
-int sys::ExecPipe::closePipe()
+namespace 
+{
+const size_t readPipe = 0;
```
asylvest: Make these `static` too.  And since they're constants, it should be `READ_PIPE`.  So, `static const size_t READ_PIPE = 0;`

modules/c++/sys/source/ExecWin32.cpp
```diff
@@ -23,10 +23,96 @@
...
+#include <sys/Exec.h>
+#include <str/Manip.h>
 
-int sys::ExecPipe::closePipe()
+#include <fcntl.h>
+#include <io.h>
+
+namespace
+{
+    const size_t readPipe = 0;
```
asylvest: Same deal in terms of renaming and `static`'ing

PR #59, opened 2015-10-13
Added math::round methods

msciaramitaro:


modules/c++/math/include/math/Round.h
```diff
@@ -0,0 +1,73 @@
...
+{
+/*!
+ *  Returns the nearest integral value towards zero.
+ *
+ *  \param value A number to evaluate
+ *  \return The 'fixed' number
+ */
+template<typename T> inline T fix(T value)
+{
+    return (value > 0.0 ? floor(value) : ceil(value));
```
asylvest: Use `std::floor()` and `std::ceil()` for consistency

modules/c++/math/include/math/Round.h
```diff
@@ -0,0 +1,73 @@
+/* =========================================================================
+ * This file is part of math-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2015, MDA Information Systems LLC
+ *
+ * types-c++ is free software; you can redistribute it and/or modify
```
asylvest: Should be `math-c++`

modules/c++/math/include/math/Round.h
```diff
@@ -0,0 +1,73 @@
...
+
+namespace math
+{
+/*!
+ *  Returns the nearest integral value towards zero.
+ *
+ *  \param value A number to evaluate
+ *  \return The 'fixed' number
+ */
+template<typename T> inline T fix(T value)
```
msciaramitaro: Is there a better name to give this?  I think *fix* is not intuitive.  Maybe something like *int* or *clip* or *roundToInt*?

asylvest: Looks like it's called `fix` because that's what Matlab calls it: http://www.mathworks.com/help/matlab/ref/fix.html.  I would suggest leaving it as-is.

PR #63, opened 2015-10-29
boost::serialize support and examples

mbojrab:
Updates to include samples for boost::serialize to disk for low level primitives. This includes unittests for the supported types. There are no waf targets for boost currently. The work is all templated so it does not require boost to build. All tests currently pass. Let me know if you want me to do something more succinct with the unittests to reduce the redundant code. There are minor differences that will make it a little tricky but doable.


mbojrab:
Design for "serialize" module is like no other library we have, so I took a first crack. This could likely be done various ways, so I made a quick module and we can modify from there. I started with having serialize method for in and out to stringstream, however this doesn't reduce the code much (if at all), so I did away with it for using the objects themselves. I assume all applications wanting this functionality will need to compile it in or use the HAVE_BOOST include guards.

asylvest:
Initial thoughts...
1. Add a modules/drivers/boost/wscript that supports a `--with-boost-home` flag.  It generates some kind of config.h like other spots do that let you know if you've got boost (without having to set a flag in the waf environment).  Don't worry about supporting building boost... just put this in to support a pre-built version and add in whatever serializer library you need as being associated with LIB_BOOST.  Or is this headers only in boost too?
2. Why do the math.linear and math.poly classes that you've added a serialize() method to actually need to include serialize/Conf.h or any boost stuff?  That serialize() method is templated, so it seems like it can always be present and doesn't require any includes.
3. Since you're not using the version right now, comment it out:

```
    template <class Archive>
    void serialize(Archive& ar, const unsigned int /*version*/)
    {
        ar & mCoef;
    }
```
this way VS won't warn about unused variables
4. In the Doxygen comments for these, maybe say 'serialize out to a boost stream' or something to make clear this is boost-specific.
5. So since we won't always have boost, and thus won't always have serializers, this means that the test_*_serialize programs won't always compile.  These would need to either be `#ifdef'd` mostly out or a fancier wscript would need to be added to only add these targets if boost is present.  I'd tend to lean towards the former because I think it'll be easier.  In this scenario then, math.linear / math.poly's DEPS wouldn't get 'serialize' added to it... instead, TEST_DEPS would.

mbojrab:
Don't merge this yet! The boost_config.h is not being picked up by the wscript for some reason (might need a second pair of eyes), and I currently assume the boost_config.h is always created however currently its only created if the user uses --with-boost-home. Once these two things are solved it should be ready. I think I've covered all your bullets above.

mbojrab:
Okay export_includes worked. I additional needed to include logic for each CRT type in boost/wscript [MT|MD|MTd|MDd]. If you would please review this version. Linux assumes the user has selected the the properly library to link against through the symlink. I think this should be acceptable logic, but comment if this is incorrect.

mbojrab:
Actually on second thought, let me rebuild on 4600a tomorrow, just to make sure Linux is still working with the new logic. I'll comment back when this is complete.

mbojrab:
Okay all tested on Linux as well. This is ready for review.

asylvest:
I added a missing newline to a file (this is my contribution to software development nowadays).

I think this looks good... I'm sure we'll tweak and iterate once we start using this and boost more extensively, but this is a good foundation to start working from.

PR #66, opened 2015-11-24
Extends SWIG bindings for math.poly

dstarinshak:
Update extends the ```__call__``` function for the three SWIG-wrapped polynomial versions: Poly1D, Poly2D, and PolyVector3. Existing functionality allows you to call a polynomial with a Python float and evaluate the polynomials at that value. New functionality allows you to call a polynomial with a Python sequence (tuple, list, ndarray, etc.) and returns a Python list of the polynomial evaluations. (For the case of Poly2D, the call is two sequences of equal length.)

Unit tests are added to verify (1) that the call functionality works and (2) that the polynomial evaluation on a sequence matches polynomial evaluation on each individual sequence entry.


modules/python/math.poly/source/math_poly.i
```diff
@@ -129,6 +147,35 @@ typedef math::linear::Vector<double> VectorDouble;
...
+        {
+            PyErr_SetString(PyExc_TypeError,"Expecting a sequence");
+            return NULL;
+        }
+        if (!PySequence_Check(y_input))
+        {
+            PyErr_SetString(PyExc_TypeError,"Expecting a sequence");
+            return NULL;
+        }
+        int N = PyObject_Length(x_input);
```
asylvest: To be pedantic, this should be a `Py_ssize_t` right?  This will make a difference on Windows if you ever had something so long it wouldn't fit in 4 bytes.  This should ripple through the for loop on line 170 and other similar spots below.

dstarinshak: Changed ‘int’ to ‘Py_ssize_t’ throughout update.

PR #67, opened 2015-12-15
Updates to net.ssl 

clydestanfield:
net.ssl now finds the openssl libraries and builds all its source by default.

There is an interface change between openssl 0.9.8(open) and openssl 1.0(back). I resolved this by building a fragment to determine which version we are using. If there is a better way to resolve this let me know.

### DIFF
--- a/modules/c++/net.ssl/wscript
+++ b/modules/c++/net.ssl/wscript
@@ -5,6 +5,7 @@ MODULE_DEPS     = 'net'
 USELIB          = 'SSL'
 
 from waflib import Options
+from build import writeConfig
 
 distclean = lambda p: None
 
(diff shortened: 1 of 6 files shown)


asylvest:
This is great.  Only catch with the interface/version check is that we want to avoid putting stuff right in `conf.define` since that'll require someone else linking this in with their own build system to make sure they define this.  Although in this case, that isn't strictly needed since it's all in a .cpp.  It would still be safer to generate a config file on the fly like we do for other modules such as sys.  I don't think we've done this for a driver before... let me know if there are any problems with doing that.

clydestanfield:
The latest version now moves the defines to a net_ssl_config.h file. I had to extract out the lib information. I didn't see any way in build.py do this automatically so I extracted it myself. Let me know if there is a better to handle this.

modules/c++/net.ssl/wscript
```diff
@@ -2,8 +2,119 @@ NAME            = 'net.ssl'
...
+                               okmsg=sslHome,
+                               mandatory=True)
+                conf.define(sslDef, True)
+            elif conf.check(lib=sslLib, uselib_store=sslUselib,
+                            header_name=sslHeader,
+                            function_name=sslFunction,
+                            fragment=sslFragment098,
+                            msg='Checking for version 0.9.8 of ssl library',
+                            mandatory=False):
+                conf.define(sslDef, True)
```
asylvest: Shouldn't only sslOldDef get set here instead of both of them?

modules/c++/net.ssl/wscript
```diff
@@ -2,8 +2,119 @@ NAME            = 'net.ssl'
...
+
+            # Update library information
+            if 'LIB_SSL' in conf.env:
+                self.uselib = conf.env['LIB_SSL']
+            if 'LIBPATH_SSL' in conf.env:
+                self.libpath = conf.env['LIBPATH_SSL']
+            if 'INCLUDES_SSL' in conf.env:
+                self.includes = conf.env['INCLUDES_SSL']
+
+        def updateConf(self, conf):
```
asylvest: I see.  Yeah this is a bummer but I'm not sure that there's a better way to do it...

modules/c++/net.ssl/wscript
clydestanfield: It does work with both of them and this way there are no #defines an external user would have to set at all. The downside is that the order of the #includes matter now within the net.ssl module. Would you rather have have USE_OPENSSL be a conf define?

PR #68, opened 2015-12-15
Warnings

JonathanMeans:
The changes to io/include/io/Dbgstream.h and re/test/PCRETest3 are more than cosmetic.


modules/c++/cli/source/ArgumentParser.cpp
```diff
@@ -249,11 +249,9 @@ cli::Results* cli::ArgumentParser::parse(const std::vector<std::string>& args)
     std::map<std::string, Argument*> longOptionsFlags;
     std::vector<Argument*> positionalArgs;
 
-    // first, validate the arguments
-    for (mem::VectorOfSharedPointers<cli::Argument>::const_iterator it =
-            mArgs.begin(); it != mArgs.end(); ++it)
+    for (const auto argument : mArgs)
```
asylvest: I should have clarified... for this particular codebase, we can't use any C++11 or newer features (yet) so this will have to get reverted back.

modules/c++/cli/unittests/test_cli.cpp
```diff
@@ -175,7 +175,7 @@ TEST_CASE(testRequired)
     TEST_ASSERT_EQ(results->get<std::string>("config"), "configFile");
 }
 
-int main(int argc, char* argv[])
+int main()
```
asylvest: While this compiles, I'd prefer the normal two argument signature, so how about
`int main(int , argc )`

modules/c++/io/include/io/DbgStream.h
```diff
@@ -63,7 +63,7 @@ class DbgStream : public OutputStream
      */
     DbgStream(OutputStream* s, bool on = false)
     {
-        set(s, false);
+        set(s, on);
```
asylvest: I agree - this looks right

modules/c++/cli/unittests/test_cli.cpp
asylvest: Same comment applies to all other `main()` changes

asylvest: FYI, You should be able to simply use git to revert these files

modules/c++/re/tests/PCRETest3.cpp
```diff
@@ -97,7 +97,7 @@ class HttpParser
 
     virtual void parse(const char* header, int length)
     {
-        mHeader = std::string(request, length);
+        mHeader = std::string(header, length);
```
asylvest: I agree - this looks right

JonathanMeans:
Fixed the signatures for main().

Not sure what's going on with sys\tests\MemoryTest1.cpp and math.poly\unittests\test_2d_poly.cpp. I took care to save them with Unix-style line endings in VS right before committing and pushing, so they should be right...

asylvest:
Looks good (we use camelCase rather than under_scores for variable names so I just fixed the `arg_it` stuff to `argIt`).  It looks like those two files did not previously have all Unix line endings, so your change there is a good one.

This is all set.

PR #69, opened 2015-12-16
Warnings

JonathanMeans:


modules/c++/xml.lite/include/xml/lite/XMLException.h
```diff
@@ -145,9 +145,15 @@ class XMLParseException : public XMLException
      *  \param col As reported by the constructor
      *
      */
-    void form(int row, int column)
+    void form(int row, int column, int errNum = 0)
     {
         std::ostringstream oss;
+	    if (errNum > 0)
```
asylvest: Looks like extra indents in this block

PR #72, opened 2015-12-28
Updated mem.i to handle auto_ptrs

chvink:
I `%template`ed the auto_ptrs in order to ignore them. I think with the `std_auto_ptr.i` included in Swig 3.0+ the `std::auto_ptr` support should be in there, but I can't think of any reason people would want to convert between `std::auto_ptr` and `mem::ScopedCopyablePtr` from the python side of things.

I also fixed some code duplication.


asylvest:
OK gotcha, so the
```
+%template(StdAuto##PyType) std::auto_ptr< namespace##::##CppType >;
```
line is the key there?

I agree that there's no need to let anyone do any smart pointer weirdness from the Python side.

asylvest:
I re-generated the files with Swig 3.0.7 to be pedantic (and caught a few others that had been generated with an older version) and fixed a typo in the .i file (`tpe` vs. `type`).  Just a typo in the macro's variable name... made no difference in the resulting generated code.

PR #73, opened 2015-12-30
Add net speed tests

asylvest:
Adding TCPClient/ServerSpeedTest applications that measure the speed of sending messages back and forth over a port via TCP


asylvest:
Made some minor tweaks to the Socket API (send/recv buffers are now `void*` rather than `sys::byte*` (no reason to make callers cast) and return a `size_t` rather than `sys::SSize_T` (if send/recv errors out and returns -1, we're going to throw inside the method anyway, so this will always be positive)

@clydestanfield @mbojrab @jmrandol @msciaramitaro

PR #75, opened 2016-01-04
calling this method with size of zero causes a segfault, because of t…

mbojrab:
…he index on line 76. xml::lite is making this call with zero at some point. Though it seems strange it would make a zero sized call, I'm fixing it here because it is ultimately a valid value.


asylvest:
I agree this is a bug.  Let's be a little more clear about what's going on though so someone doesn't think this is an extraneous check in the future.  What about:

```
// Guard against &mData[mPosition] below indexing out of bounds
// when mPosition is still 0
if (size > 0)
{
    sys::Size_T newPos = mPosition + size;
    if (newPos >= mData.size())
        mData.resize(newPos);
    std::copy(b, b+size, &mData[mPosition]);
    mPosition = newPos;
}
```

mbojrab:
I made the update, I agree this is more intuitive.

PR #76, opened 2016-01-08
Fixing the broken subprocess. This was unable to call shell commands like cd, as it was before

mbojrab:
This is a pretty straigh-foward update. Not sure why I chose the other route as this was what popen is doing under the hood. Easy-Peasy.


asylvest:
@msciaramitaro

PR #90, opened 2016-03-09
Make equality operators non-member functions

JonathanMeans:
This causes SWIG to not wrap the operator.


asylvest:
Solved the problem downstream rather than here

PR #91, opened 2016-03-15
Use waitpid to reap status of forked processes and avoid defunct states

msciaramitaro:


asylvest:
Is it possible to wait until @mbojrab is back for him to weigh in or do you need this immediately?

msciaramitaro:
We need to start testing AutoIMX soon, but I guess we can merge it in mid-testing since it doesn't actually break processing.

asylvest:
If it gets to the point where you need it, I'm fine merging this in and having Micah review when he returns... if you can wait until next week, @mbojrab can you review it as soon as you can upon your return?

asylvest:
Discussed internally - believe latest changes reflect what we want

PR #92, opened 2016-03-21
Added isSetEnv and getEnvIfSet

apakanati:
Picked up some non-code changes  e.g. OSUnix.cpp:257-260 -> 281-284.  Not really sure why, but nothing should have changed in there.

Built on Win & Unix since there's define guards that change compilation on OS type.  Any testing recommendations welcome too.  Didn't see any tests pertaining to env var in the code base.  Could add a couple, but all I can reasonably think of is:  set -> isenvset -> get .   Let me know what you think here.

Also I added some more duplication of code between Win32 and Unix.  I chose to kept the abstract base class empty, but it might be better from a maintenance standpoint to put that in the base class, so feedback there would also be useful.

### DIFF
--- a/modules/c++/sys/include/sys/AbstractOS.h
+++ b/modules/c++/sys/include/sys/AbstractOS.h
@@ -192,9 +192,10 @@ public:
     virtual bool isEnvSet(const std::string& s) const = 0;
 
     /*!
-     *  Get an environment variable, if set
+     *  Get an environment variable and updates value, but only if set.
+     *  Returns true if environment variable is set, false otherwise
      */
-    virtual bool getEnvIfSet(const std::string& envVar, std::string& value) const = 0;
+    bool getEnvIfSet(const std::string& envVar, std::string& value) const;
 
     /*!
      *  Set an environment variable
--- a/modules/c++/sys/include/sys/AbstractOS.h
+++ b/modules/c++/sys/include/sys/AbstractOS.h
@@ -187,6 +187,16 @@ public:
     virtual std::string getEnv(const std::string& s) const = 0;
 
     /*!
+     *  Returns true if environment variable is set, false otherwise
+     */
+    virtual bool isEnvSet(const std::string& s) const = 0;
+
+    /*!
+     *  Get an environment variable, if set
+     */
+    virtual bool getEnvIfSet(const std::string& envVar, std::string& value) const = 0;
+
+    /*!
      *  Set an environment variable
      */
     virtual void setEnv(const std::string& var, 
--- a/modules/c++/sys/include/sys/OSUnix.h
+++ b/modules/c++/sys/include/sys/OSUnix.h
@@ -143,6 +143,17 @@ public:
     virtual std::string getEnv(const std::string& s) const;
 
     /*!
+     * Returns true if environment variable is set, false otherwise
+     */
+    virtual bool isEnvSet(const std::string& s) const;
+
+    /*!
+     * Gets environment variable, if set
+     */
+
+    virtual bool getEnvIfSet(const std::string& envVar, std::string& value) const;
+
+    /*!
      *  Set an environment variable
      */
     virtual void setEnv(const std::string& var, 
--- a/modules/c++/sys/include/sys/OSUnix.h
+++ b/modules/c++/sys/include/sys/OSUnix.h
@@ -148,12 +148,6 @@ public:
     virtual bool isEnvSet(const std::string& s) const;
 
     /*!
-     * Gets environment variable, if set
-     */
-
-    virtual bool getEnvIfSet(const std::string& envVar, std::string& value) const;
-
-    /*!
      *  Set an environment variable
      */
     virtual void setEnv(const std::string& var, 
--- a/modules/c++/sys/source/OSUnix.cpp
+++ b/modules/c++/sys/source/OSUnix.cpp
@@ -210,6 +210,30 @@ std::string sys::OSUnix::getEnv(const std::string& s) const
     return std::string(p); 
 }
 
+bool sys::OSUnix::isEnvSet(const std::string& s) const
+{
+    try
+    {
+        getEnv(s);
+    }
+    catch(sys::SystemException)
+    {
+        return false;
+    }
+    return true;
+}
+
+bool sys::OSUnix::getEnvIfSet(const std::string& envVar, std::string& value) const
+{
+    if(isEnvSet(envVar))
+    {
+        value = getEnv(value);
+        return true;
+    }
+    return false;
+}
+
+
 void sys::OSUnix::setEnv(const std::string& var, 
                          const std::string& val,
                          bool overwrite)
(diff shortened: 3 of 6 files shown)


modules/c++/sys/include/sys/AbstractOS.h
```diff
@@ -187,6 +187,16 @@ class AbstractOS
     virtual std::string getEnv(const std::string& s) const = 0;
 
     /*!
+     *  Returns true if environment variable is set, false otherwise
+     */
+    virtual bool isEnvSet(const std::string& s) const = 0;
+
+    /*!
+     *  Get an environment variable, if set
```
asylvest: Add some details about how returns true/false based on if env variable is set and what `value` is set to in the two cases

modules/c++/sys/include/sys/OSUnix.h
```diff
@@ -143,6 +143,17 @@ class OSUnix : public AbstractOS
...
 
     /*!
+     * Returns true if environment variable is set, false otherwise
+     */
+    virtual bool isEnvSet(const std::string& s) const;
+
+    /*!
+     * Gets environment variable, if set
+     */
+
```
asylvest: Add newline

asylvest: Err rather there's an extra newline that should be removed

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -210,6 +210,30 @@ std::string sys::OSUnix::getEnv(const std::string& s) const
     return std::string(p); 
 }
 
+bool sys::OSUnix::isEnvSet(const std::string& s) const
```
asylvest: Rather than a try/catch, I'd call the underlying `getenv()` directly

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -210,6 +210,30 @@ std::string sys::OSUnix::getEnv(const std::string& s) const
...
+    catch(sys::SystemException)
+    {
+        return false;
+    }
+    return true;
+}
+
+bool sys::OSUnix::getEnvIfSet(const std::string& envVar, std::string& value) const
+{
+    if(isEnvSet(envVar))
```
asylvest: Extra space: `if (isEnvSet(envVar))`

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -229,6 +229,29 @@ std::string sys::OSWin32::getEnv(const std::string& s) const
     return result;
 }
 
+bool sys::OSWin32::isEnvSet(const std::string& s) const
+{
+    try
```
asylvest: Same deal - avoid the try/catch

asylvest: And as long as you're in the code... In the `getEnv()` implementation, can you switch this to a vector rather than a raw pointer?

modules/c++/sys/include/sys/AbstractOS.h
```diff
@@ -187,6 +187,16 @@ class AbstractOS
...
+     */
+    virtual bool getEnvIfSet(const std::string& envVar, std::string& value) const = 0;
```
asylvest: This can just be implemented in the base class just once right?

apakanati:
Changes made, I think.  Please see 225-228 of OSWin32.cpp replacing the raw pointer allocation.  I'm not sure if this is actually an improvement over the original, but open to suggestions if you had something else in mind..

+    std::vector<char> buffer;
+    buffer.resize(size+1);
+    DWORD retVal = GetEnvironmentVariable(s.c_str(), static_cast<char*>(&buffer[0]), size);
+    result = static_cast<char*>(&buffer[0]);

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -210,6 +210,14 @@ std::string sys::OSUnix::getEnv(const std::string& s) const
     return std::string(p); 
 }
 
+bool sys::OSUnix::isEnvSet(const std::string& s) const
+{
+    const char* p = getenv(s.c_str());
+    if (p == NULL)
```
asylvest: Try to avoid single character variable names (though I'm guessing getEnv() did this).  Also can simplify the return stuff to a single line.  How about...

```
const char* const envVal  = getenv(s.c_str());
return (envVal != NULL);
```

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -221,14 +222,21 @@ std::string sys::OSWin32::getEnv(const std::string& s) const
 
     // If we can use a normal size buffer, lets not bother to malloc
 
-    char *buffer = new char[size + 1];
-    DWORD retVal = GetEnvironmentVariable(s.c_str(), buffer, size);
-    result = buffer;
-    delete [] buffer;
-
+    std::vector<char> buffer;
```
asylvest: Just do this in one shot: `std::vector<char> buffer(size + 1);`
(Also note spaces between the operators there)

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -221,14 +222,21 @@ std::string sys::OSWin32::getEnv(const std::string& s) const
...
+    buffer.resize(size+1);
+    DWORD retVal = GetEnvironmentVariable(s.c_str(), static_cast<char*>(&buffer[0]), size);
```
asylvest: Don't need the `static_cast` here or in the next line... `&buffer[0]` is a `char*`

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -221,14 +222,21 @@ std::string sys::OSWin32::getEnv(const std::string& s) const
 
     // If we can use a normal size buffer, lets not bother to malloc
```
asylvest: Not sure what this comment was ever for but doesn't seem to apply so remove it

modules/c++/sys/source/OSWin32.cpp
asylvest: To be pedantic, add a check that `retVal == size` and throw if it doesn't (this would only happen if someone else is setting the environment variable to a different value between the two checks... that would be a bad day)

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -221,14 +222,21 @@ std::string sys::OSWin32::getEnv(const std::string& s) const
...
+    buffer.resize(size+1);
+    DWORD retVal = GetEnvironmentVariable(s.c_str(), static_cast<char*>(&buffer[0]), size);
+    result = static_cast<char*>(&buffer[0]);
     return result;
 }
 
+bool sys::OSWin32::isEnvSet(const std::string& s) const
+{
+    DWORD size = GetEnvironmentVariable(s.c_str(), NULL, 0);
+    if (size == 0)
```
asylvest: Just do `return (size != 0);`

apakanati:
Updated again

PR #93, opened 2016-03-30
Fixing incorrect call to getenv

apakanati:
Well, oops.

@asylvest Should I just commit this as is in orcrist too or will that create more problems?


apakanati:
Closing this request in favor of another one.

PR #95, opened 2016-04-06
sys::SemaphoreWin32 max count data type 

jbruce2112:
`sys::SemaphoreWin32` has a max count data type of `LONG` which is a valid Windows #define. However, some evil libraries define that or undefine it for themselves, so we should only reference it in the .cpp file not the .h file. Modify the class definition and .h file to use a `size_t` for max count, then cast to `LONG` in the .cpp file.

### DIFF
--- a/modules/c++/sys/source/SemaphoreWin32.cpp
+++ b/modules/c++/sys/source/SemaphoreWin32.cpp
@@ -27,12 +27,15 @@
 
 #include "sys/SemaphoreWin32.h"
 
-sys::SemaphoreWin32::SemaphoreWin32(unsigned int count, LONG maxCount)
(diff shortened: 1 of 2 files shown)


modules/c++/sys/source/SemaphoreWin32.cpp
```diff
@@ -27,12 +27,15 @@
 
 #include "sys/SemaphoreWin32.h"
 
-sys::SemaphoreWin32::SemaphoreWin32(unsigned int count, LONG maxCount)
+sys::SemaphoreWin32::SemaphoreWin32(unsigned int count, size_t _maxCount)
 {
+	// Ensure maxCount never becomes negative due to casting between signed/unsigned types
```
jmrandol: CODA standards (somewhere?) indicate a 4 space indent

PR #96, opened 2016-04-06
Remove plugin manager warning

jbruce2112:
When the plugin manager finds a DLL or SO that doesn't fit the "plugin" mold, it prints out a warning to stdout or stderr "Warning: plugin manager failed to load: Failed to load function: GetPluginIdentity: The specified procedure could not be found"

I added an optional logger to `DefaultErrorHandler` that will pass all messages to the logger rather than cout when specified. The plugin module now has a dependency on the logging module because of this.

### DIFF
--- a/modules/c++/plugin/include/plugin/ErrorHandler.h
+++ b/modules/c++/plugin/include/plugin/ErrorHandler.h
@@ -25,6 +25,7 @@
 
 #include <import/except.h>
 #include <import/sys.h>
+#include <import/logging.h>
 
 namespace plugin
 {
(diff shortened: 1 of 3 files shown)


modules/c++/plugin/include/plugin/ErrorHandler.h
```diff
@@ -44,8 +45,7 @@ class ErrorHandler
 class DefaultErrorHandler : public ErrorHandler
 {
 public:
-    DefaultErrorHandler();
-    virtual ~DefaultErrorHandler();
+    DefaultErrorHandler(logging::Logger* = NULL);
```
jmrandol: CODA standards (somewhere?) desire to name variables in function signatures

PR #100, opened 2016-04-11
Make the logger member of DefaultErrorHandler a SharedPtr & reduce some log levels.

jbruce2112:
As discussed with @msciaramitaro and @asylvest, removing `TaskErrorHandler` broke some behaviors others relied on. This fixes some of it.

### DIFF
--- a/modules/c++/plugin/source/ErrorHandler.cpp
+++ b/modules/c++/plugin/source/ErrorHandler.cpp
@@ -23,7 +23,7 @@
 #include "plugin/ErrorHandler.h"
 
 plugin::DefaultErrorHandler::DefaultErrorHandler(logging::Logger* logger) :
-	mLogger(logger)
+	mLogger(mem::SharedPtr<logging::Logger>(logger))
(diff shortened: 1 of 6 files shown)


modules/c++/plugin/source/ErrorHandler.cpp
```diff
@@ -23,7 +23,7 @@
 #include "plugin/ErrorHandler.h"
 
 plugin::DefaultErrorHandler::DefaultErrorHandler(logging::Logger* logger) :
-	mLogger(logger)
+	mLogger(mem::SharedPtr<logging::Logger>(logger))
```
jmrandol: This seems bad.  If we're going to use `SharedPtr` it needs to be passed in that way as well, otherwise this doesn't buy us anything and will likely blow up.

jbruce2112: Agreed. Just pushed a change to make them `SharedPtr`s throughout.

PR #101, opened 2016-04-12
Inf equality

JonathanMeans:
Previously, equals(infinity, infinity) returned false.

### DIFF
--- a/modules/c++/math.linear/include/math/linear/MatrixMxN.h
+++ b/modules/c++/math.linear/include/math/linear/MatrixMxN.h
@@ -38,35 +38,23 @@ template<typename _T> bool equals(const _T& e1, const _T& e2)
 
 template<typename _T> inline bool equals(const _T& e1, const _T& e2, _T eps)
 {
+    if (std::numeric_limits<_T>::has_infinity)
+    {
+        if (e1 == std::numeric_limits<_T>::infinity() ||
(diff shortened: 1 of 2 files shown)


modules/c++/math.linear/include/math/linear/MatrixMxN.h
```diff
@@ -43,10 +43,30 @@ template<typename _T> inline bool equals(const _T& e1, const _T& e2, _T eps)
 
 template<> inline bool equals(const float& e1, const float& e2)
 {
+    if (e1 == std::numeric_limits<float>::infinity() &&
```
asylvest: Rather than this, shouldn't the base `equals()` method that takes in an `eps` be updated to do this?

To confirm, if you say
```
float e1 = std::numeric_limits<float>::infinity();
float e2 = std::numeric_limits<float>::infinity();
bool val =  e1 == e2;
```

`val` will be true right?  If so, I agree that we should add this logic.

JonathanMeans: Fixed, and I’ve confirmed that equality for numeric_limits<float>::infinity() works like that on both Linux and Windows.

asylvest:
Cool.  Last thing... include `<limits`>

JonathanMeans:
Done.

PR #102, opened 2016-04-13
These setupLogger() functions should have kept their auto_ptr return type

jbruce2112:


asylvest:
Thanks.  Should the `LoggerPtr` typedef in `Logger.h` be removed too then?

jbruce2112:
We actually still use that typedef in `plugin::DefaultErrorHandler` and in a few other areas in `common.task.*`

asylvest:
I am not sold on that that typedef really belongs in here since the logging module doesn't actually use it  now but I guess it doesn't really matter :o)

Issue #103, opened 2016-04-13
Add mem::SharedPtr::reset(std::auto_ptr<T> ptr)


asylvest:
We have a constructor that takes in an `auto_ptr` but not a `reset()` method that does... add it.

@jmrandol @jbruce2112

PR #105, opened 2016-04-18
Add a reset() overload for mem::SharedPtr which takes an auto_ptr

jbruce2112:
This is for issue #103 

I added a overload for `mem::SharedPtr<T>::reset` that takes in an `std::auto_ptr<T>` and did a bit of refactoring so we could use the existing code for `reset()`. Also added a unit test.

### DIFF
--- a/modules/c++/mem/include/mem/SharedPtr.h
+++ b/modules/c++/mem/include/mem/SharedPtr.h
@@ -144,17 +144,14 @@ public:
         return mRefCtr->get();
     }
 
-    void reset(T* ptr = NULL)
+    void reset(std::auto_ptr<T> scopedPtr)
     {
-        // We take ownership of the pointer no matter what, so temporarily
-        // wrap it in an auto_ptr in case creating the atomic counter throws.
         // NOTE: We need to create newRefCtr on the side before decrementing
         //       mRefCtr. This way, we can provide the strong exception
         //       guarantee (i.e. the operation either succeeds or throws - the
         //       underlying object is always in a good state).
-        std::auto_ptr<T> scopedPtr(ptr);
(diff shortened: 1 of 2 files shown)


modules/c++/mem/include/mem/SharedPtr.h
```diff
@@ -168,6 +165,14 @@ class SharedPtr
         mPtr = ptr;
     }
 
+    void reset(T* ptr = NULL)
+    {
+        // We take ownership of the pointer no matter what, so
+        // temporarily wrap it in an auto_ptr in case creating
+        // the atomic counter throws later on.
```
asylvest: Change 'later on' to 'in the underlying method'

modules/c++/mem/include/mem/SharedPtr.h
```diff
@@ -144,17 +144,14 @@ class SharedPtr
...
-        // We take ownership of the pointer no matter what, so temporarily
-        // wrap it in an auto_ptr in case creating the atomic counter throws.
         // NOTE: We need to create newRefCtr on the side before decrementing
         //       mRefCtr. This way, we can provide the strong exception
         //       guarantee (i.e. the operation either succeeds or throws - the
         //       underlying object is always in a good state).
-        std::auto_ptr<T> scopedPtr(ptr);
         sys::AtomicCounter* const newRefCtr = new sys::AtomicCounter(1);
-        scopedPtr.release();
+        T* ptr = scopedPtr.release();
```
asylvest: Rather than doing this, you can say `mPtr = scopedPtr.release()` below right?  That makes really explicit the point at which ownership is being handed over.

PR #106, opened 2016-04-19
Update build scripts to be dual-compatible with Python 2 and 3

JonathanMeans:
I made an assumption that dir names would not require Unicode support.


asylvest:
@jmrandol

PR #107, opened 2016-04-21
Dual compatability

JonathanMeans:


asylvest:
OK cool.  So @JonathanMeans  if you build with Swig with the `-py3` flag, are the C++ and Python files it generates such that they'll also work with Python 2?  So going forward whenever we update any C++ API stuff and have the resulting Swig-generated files change, we should make sure we always do this with Python 3?  Or are files with -py3 not compatible with Python 2?

JonathanMeans:
The files with –py3 files are not compatible with Python 2 due to some changes in Python’s C API between versions.

asylvest:
OK bummer.  So are the .cxx files that were changed under this pull request generated with `-py3` or what is causing those diffs?  I think we'll want to just check in files compatible with Python 2 in the immediate term (but now we allow people to generate Python 3-compatible code which is great).

JonathanMeans:
So the diffs in the .cxx files are preprocessor conditionals that will generate different code for Python 2 vs 3 at compile time.

Although looking at the documentation more closely, it’s possible the –py3 flag just enables optional/convenience features? I’ll test without that and see what happens.

PR #110, opened 2016-04-22
mem::VectorOfPointers<>::erase

jbruce2112:
As I'm going through some SDS code, I found a similar class in that codebase to our `mem::VectorOfPointers`, so I removed it and added the `erase()` function to CODA's so we can instead use the CODA version.

### DIFF
--- a/modules/c++/mem/include/mem/VectorOfPointers.h
+++ b/modules/c++/mem/include/mem/VectorOfPointers.h
@@ -26,6 +26,7 @@
 #include <cstddef>
 #include <vector>
 #include <memory>
+#include <algorithm>
 
 #include <mem/SharedPtr.h>
 
@@ -122,19 +123,7 @@ public:
         return mValues.erase(pos);
     }
 
-    const_iterator erase(const const_iterator &pos)
-    {
-        delete *pos;
-        return mValues.erase(pos);
-    }
-
     iterator erase(const iterator &first, const iterator &last)
-    {
-        std::for_each(first, last, Deleter());
-        return mValues.erase(first, last);
-    }
-
-    const_iterator erase(const const_iterator &first, const const_iterator &last)
     {
         std::for_each(first, last, Deleter());
         return mValues.erase(first, last);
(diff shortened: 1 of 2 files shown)


modules/c++/mem/include/mem/VectorOfPointers.h
```diff
@@ -116,10 +117,24 @@ class VectorOfPointers
     iterator end() { return mValues.end(); }
     const_iterator end() const { return mValues.end(); }
 
+    iterator erase(const iterator &pos)
```
asylvest: Iterators can be passed in by value rather than const ref (they're passed in by value throughout the STL)

modules/c++/mem/include/mem/VectorOfPointers.h
```diff
@@ -116,10 +117,24 @@ class VectorOfPointers
...
+    {
+        delete *pos;
+        return mValues.erase(pos);
+    }
+
+    iterator erase(const iterator &first, const iterator &last)
+    {
+        std::for_each(first, last, Deleter());
```
asylvest: So this is a style/opinion thing... If you could use a lambda here it would save you a few lines and `for_each()` would buy you something.  With something this simple, granted the name `Deleter` is obvious, but I have to go look at the struct to make sure and I think it's simpler code to skim over by just using a for loop.  Style-wise, we'll want curly braces even for one liner for loops FYI.

jbruce2112: I agree. Changing it to a manual loop in this case is a little more readable.

modules/c++/mem/include/mem/VectorOfPointers.h
```diff
@@ -26,6 +26,7 @@
 #include <cstddef>
 #include <vector>
 #include <memory>
+#include <algorithm>
```
asylvest: Can be removed now

modules/c++/mem/include/mem/VectorOfPointers.h
```diff
@@ -116,6 +117,24 @@ class VectorOfPointers
...
+        return mValues.erase(pos);
+    }
+
+    iterator erase(iterator first, iterator last)
+    {
+        iterator iter = first;
+        while (iter != last)
+        {
+            delete *iter;
+            iter++;
```
asylvest: Always `++iter` just in case `iter++` is expensive

PR #116, opened 2016-05-04
Fix Solaris build by disabling Python and replacing deprecated os.pop…

JonathanMeans:
…en with subprocess.Popen

### DIFF
--- a/build/pythontool.py
+++ b/build/pythontool.py
@@ -4,11 +4,12 @@ import os, sys, re
 def options(opt):
     opt.load('python')
     if sys.platform == 'sunos5':
-        opt.add_option('--disable-python', action='store_true', 
+        opt.add_option('--enable-python', action='store_true',
                 dest='python', help='Disable python', default=False)
     else:
-        opt.add_option('--disable-python', action='store_false', dest='python',
-                   help='Disable python', default=True)
+        opt.add_option('--disable-python', action='store_false',
+                dest='python', help='Disable python', default=True)
+
     opt.add_option('--python-version', action='store', dest='python_version',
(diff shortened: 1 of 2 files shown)


build/pythontool.py
```diff
@@ -1,9 +1,13 @@
 from waflib import Options
-import os, re
+import os, sys, re
 
 def options(opt):
     opt.load('python')
-    opt.add_option('--disable-python', action='store_false', dest='python',
+    if sys.platform == 'sunos5':
+        opt.add_option('--disable-python', action='store_true',
```
asylvest: Actually I misspoke earlier... let's make this one `--enable-python`.  Otherwise they don't have a way to actually turn Python support on in Solaris do they?

JonathanMeans: Ah, right. Fixed.

build/pythontool.py
```diff
@@ -1,10 +1,15 @@
...
-import os, re
+import os, sys, re
 
 def options(opt):
     opt.load('python')
-    opt.add_option('--disable-python', action='store_false', dest='python',
-                   help='Disable python', default=True)
+    if sys.platform == 'sunos5':
+        opt.add_option('--enable-python', action='store_true',
+                dest='python', help='Disable python', default=False)
```
asylvest: help should be "Enable python"

asylvest: Add a comment for why we're doing this... something like "The Solaris Studio compiler cannot compile some of the C++ bindings that Swig produces so default to no Python bindings on Solaris so the build doesn't fail"

JonathanMeans: Done.

PR #122, opened 2016-05-09
Numpyutils update2

apakanati:
Take a look at the #define stuff I added in the .h.  I'm not really sure this was a good idea, but as a brief explanation of what's going on.

- we're including numpy headers in the numpyutils header
- this means that all downstream compilation units get a warning because they don't have their own import_arrays call
- I tried some gymnastics with forward declaring numpy stuff, but ran into some trouble with C and typedef forwards and gave up on the basis of not being worth the time.  Not sure if this is possible but this might be an a good alternative (and would have the benefit of not appearing to provide numpy operators to downstream modules) when we don't safely

So I settled on this to #define NO_IMPORT which causes the header to suppress those warnings (and the definitions of import_arrays).  And then added some other stuff just in case NO_IMPORT is 
set beforehand in which case I don't want to clear it.

We could also just accept the warning and not do any of this of course.

### DIFF
--- a/modules/c++/numpyutils/examples/scripts/numpyutilstest.py
+++ b/modules/c++/numpyutils/examples/scripts/numpyutilstest.py
@@ -29,7 +29,6 @@ import numpy
 # SWIG interface.  This is the python version calling the C implementation
 # in source/numpyutilstest.i
 
-
 print(coda.numpyutilstest.elementDoubleTest.__doc__)
 
 numpyarray = numpy.zeros((2,3), dtype=numpy.float32)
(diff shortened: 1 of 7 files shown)


asylvest:
Bah.  So I am torn... in reality Swig-generated code generates lots of compiler warnings and there's nothing we can do about that, so it's going to be just life that the Python bindings will have annoying warnings.  I do think the NPY_NO_DEPRECATED_API is worth putting in to remove that warning, but I'd say remove the NO_IMPORT defines and we'll just live with that warning.  I see what that logic is doing and it seems like it should work, but it just feels dirty.  We can always keep it in our back pocket or keep thinking about if there's a cleaner way to do this.

apakanati:
Removed the defines as per your recommendation.  It also occurred to me that PyArray_BYTES might cause issues, so I decided to be paranoid.

Issue #123, opened 2016-05-11
numpyutils fixes


asylvest:
1. modules/drivers/numpy/wscript: Should honor `--disable-python`
2. modules/drivers/numpy/wscript: Should check for `numpy/arrayobject.h` instead... just ran into a weird case where with Anaconda 2.1, compiling this module failed due to it wanting Unicode support (not sure if this is Anaconda version-specific or someone did a weird Anaconda installation).  Either way, we should have booted out NumPy at configure time.
3. modules/c++/numpyutils/wscript: Doesn't need `Options.options.enable_numpy`

PR #124, opened 2016-05-11
Made numpy driver more robust

chvink:
I missed this when I moved this code into coda-oss. Jeff Pursell found the problem earlier today.

1. Added check to make sure python is enabled.
2. conf.check won't throw unless mandatory is true, getting a false positive if numpy/python were missing.
3. The msg() was unnecessary.

### DIFF
--- a/modules/drivers/numpy/wscript
+++ b/modules/drivers/numpy/wscript
@@ -10,7 +10,7 @@ def options(opt):
                    help='Turn off Numpy C headers/libs')
 
 def configure(conf):    
-    if Options.options.enable_numpy:
+    if Options.options.enable_numpy and Options.options.python:
         # a little magic borrowed from python-config and extended for numpy and -L flags
         try:
             from distutils import sysconfig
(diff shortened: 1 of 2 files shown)


asylvest:
Ahh Jeff talked to both of us. :o)

Take a look at #123 for a few other things I just put out there

@apakanati

modules/drivers/numpy/wscript
```diff
@@ -34,10 +34,9 @@ def configure(conf):
                        includes=inc_dir,
                        libpath=lib_dir,
                        msg='Checking for library numpy' + warningmessage, 
-                       mandatory=False) 
-            
+                       mandatory=True)
```
asylvest: It's not mandatory True vs. False so much as that you have to check for a different header right?

chvink: Hmm... When configuring and looking for `numpy/arrayobject.h` waf never finds anything for me on miinfcompute01. With mandatory=False configuration completes successfully and compilation fails when it can't find `numpy/arrayobject.h`. With mandatory=True configuration completes successfully and the numpyutils-c++ task never gets created. But when looking for "numpy/numpyconfig.h" waf _does_ find it and successfully compiles (regardless of mandatory setting)?

asylvest: `numpy/arrayobject.h` is what `numpyutils.h` uses... it must be somewhere :o)

chvink: Fixed, it just had to bring in PYEXT when numpyconfig.h didn't. I still think mandatory needs to be true to stop PY_HAVE_NUMPY from getting defined when it's not supposed to, though.

apakanati:
Sorry about this-- managed to miss this.

A note about arrayobject.h  : since Clyde and I spent some time staring at this issue and it took a while to figure it out.

Since it compiles a test harness, not all the H files are equally usable since some require python headers and other stuff (that's my current theory).  That's why I ended up picking that obscure file.

PR #125, opened 2016-05-12
Center input values being used for 2D poly fit

msciaramitaro:
Center input values for 2D polynomial fit.

The unit test has been updated with a more stressing test using more points and a higher order polynomial.  The previous version of math::poly::fit would fail this test with a mean residual error of 4.96674e+12.  The new version achieves a mean residual error of 1.39047e-14.

### DIFF
--- a/modules/c++/math.poly/include/math/poly/Fit.h
+++ b/modules/c++/math.poly/include/math/poly/Fit.h
@@ -5,6 +5,7 @@
 #include <math/poly/TwoD.h>
 #include <math/linear/Matrix2D.h>
 #include <math/linear/VectorN.h>
+#include <numeric>
 
 namespace math
 {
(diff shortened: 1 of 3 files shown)


modules/c++/math.poly/include/math/poly/Fit.h
```diff
@@ -114,30 +114,63 @@ inline math::poly::TwoD<double> fit(const math::linear::Matrix2D<double>& x,
...
+    size_t mxn = x.size();
+
     if (m != y.rows())
         throw except::Exception(Ctxt("Matrices must be equally sized"));
 
     if (n != y.cols())
         throw except::Exception(Ctxt("Matrices must be equally sized"));
-    
+
+    // Compute min and max values for centering
```
asylvest: Since x and y are the same, this code can be cut in half if you put in a helper function that takes in the original matrix and provides back the centered matrix.  While you're at it, you could move the normalization and scalar division (which were already in the code and are also duplicated for x and y) into that function too.

modules/c++/math.poly/include/math/poly/Fit.h
```diff
@@ -233,9 +263,15 @@ inline math::poly::TwoD<double> fit(const math::linear::Matrix2D<double>& x,
         }
         xacc *= rxrms;
     }
-    return coeffs;
 
+    math::poly::TwoD<double> shift_x(1, 1);
```
asylvest: Add a comment for what's happening here.  Also, variables should be camelCase.

modules/c++/math.poly/unittests/test_llsq.cpp
```diff
@@ -90,9 +90,82 @@ TEST_CASE(test2DPolyfit)
     TEST_ASSERT_EQ(poly, truth);
 }
 
+TEST_CASE(test2DPolyfitLarge)
+{
+    // Use a defined polynomial to generate mapped values.  This ensures
+    // it is possible to fit the points using at least as many coefficients.
+    double coeffs[] =
```
asylvest: Can be `const`

msciaramitaro:
Added a minMax() method to Matrix2D.  None of the other options were really viable since intermediate values are needed later in the fit call.

Note the matrix normalization is not the same as Matrix2D's norm().  It has an extra divide by number of elements.  I tried using norm() directly and it resulted in a non-invertible matrix.  I did update to use normSq() though.

msciaramitaro:
Note to update swig if applicable.

PR #128, opened 2016-05-14
Center and normalize 1d poly fit inputs similar to 2d fit.

msciaramitaro:


asylvest:
I added a unit test - this is good to go

PR #133, opened 2016-06-01
fix (or turn off) some compiler warnings

JDanielSmith:
Incremental progress towards fixing (or turning off) various compiler warnings.  Some of these are rather pedantic (to say the least), but they can hide other (more important) warnings.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -969,9 +969,10 @@ def configureCompilerOptions(self):
         # specifications properly.
         # Skipping warning 4512 about being unable to generate an assignment
         # operator (since we often do this intentionally).
+	# warning C4100: '...' : unreferenced formal parameter
         # For warnings, use /W4 because /Wall
         # gives us tons of warnings in the VS headers themselves
-        warningFlags = '/W4 /wd4290 /wd4512'
+        warningFlags = '/W4 /wd4290 /wd4512 /wd4100'
         if Options.options.warningsAsErrors:
             warningFlags += ' /WX'
 
--- a/build/build.py
+++ b/build/build.py
@@ -969,10 +969,11 @@ def configureCompilerOptions(self):
         # specifications properly.
         # Skipping warning 4512 about being unable to generate an assignment
         # operator (since we often do this intentionally).
-	# warning C4100: '...' : unreferenced formal parameter
+        # warning C4100: '...' : unreferenced formal parameter
+        # warning C4127: conditional expression is constant
         # For warnings, use /W4 because /Wall
         # gives us tons of warnings in the VS headers themselves
-        warningFlags = '/W4 /wd4290 /wd4512 /wd4100'
+        warningFlags = '/W4 /wd4290 /wd4512 /wd4100 /wd4127'
         if Options.options.warningsAsErrors:
             warningFlags += ' /WX'
 
--- a/modules/c++/sys/source/OSWin32.cpp
+++ b/modules/c++/sys/source/OSWin32.cpp
@@ -100,7 +100,7 @@ sys::Pid_T sys::OSWin32::getProcessId() const
 
 void sys::OSWin32::removeFile(const std::string& pathname) const
 {
-    if (DeleteFile(pathname.c_str()) != TRUE)
+    if (DeleteFile(pathname.c_str()) == FALSE) // "If the function succeeds, the return value is nonzero. If the function fails, the return value is zero (0)."
     {
         sys::Err err;
         std::ostringstream oss;
@@ -113,7 +113,7 @@ void sys::OSWin32::removeFile(const std::string& pathname) const
 
 void sys::OSWin32::removeDirectory(const std::string& pathname) const
 {
-	if (RemoveDirectory(pathname.c_str()) != TRUE)
+    if (RemoveDirectory(pathname.c_str()) == FALSE) // "If the function succeeds, the return value is nonzero. If the function fails, the return value is zero (0)."
     {
         sys::Err err;
         std::ostringstream oss;
@@ -312,7 +312,7 @@ void sys::OSWin32::createSymlink(const std::string& origPathname,
 
 void sys::OSWin32::removeSymlink(const std::string& symlinkPathname) const
 {
-	if (RemoveDirectory(symlinkPathname.c_str()) != TRUE)
+	if (RemoveDirectory(symlinkPathname.c_str()) == FALSE) // "If the function succeeds, the return value is nonzero. If the function fails, the return value is zero (0)."
     {
         sys::Err err;
         std::ostringstream oss;
--- a/modules/c++/sys/source/OSWin32.cpp
+++ b/modules/c++/sys/source/OSWin32.cpp
@@ -100,7 +100,7 @@ sys::Pid_T sys::OSWin32::getProcessId() const
 
 void sys::OSWin32::removeFile(const std::string& pathname) const
 {
-    if (DeleteFile(pathname.c_str()) != true)
+    if (DeleteFile(pathname.c_str()) != TRUE)
     {
         sys::Err err;
         std::ostringstream oss;
@@ -113,7 +113,7 @@ void sys::OSWin32::removeFile(const std::string& pathname) const
 
 void sys::OSWin32::removeDirectory(const std::string& pathname) const
 {
-    if (RemoveDirectory(pathname.c_str()) != true)
+	if (RemoveDirectory(pathname.c_str()) != TRUE)
     {
         sys::Err err;
         std::ostringstream oss;
@@ -312,7 +312,7 @@ void sys::OSWin32::createSymlink(const std::string& origPathname,
 
 void sys::OSWin32::removeSymlink(const std::string& symlinkPathname) const
 {
-	if (RemoveDirectory(symlinkPathname.c_str()) != true)
+	if (RemoveDirectory(symlinkPathname.c_str()) != TRUE)
     {
         sys::Err err;
         std::ostringstream oss;
--- a/modules/c++/sys/unittests/test_symlink.cpp
+++ b/modules/c++/sys/unittests/test_symlink.cpp
@@ -33,8 +33,6 @@ namespace
 
 TEST_CASE(testCreateSymlink)
 {
-	bool test_assert_false = false;
-
     try
     {
         sys::OS os;
@@ -69,18 +67,18 @@ TEST_CASE(testCreateSymlink)
     catch (const std::exception& ex)
     {
         std::cerr << "Caught std::exception: " << ex.what() << std::endl;
-		TEST_ASSERT(test_assert_false);
+        TEST_ASSERT(false);
     }
     catch (const except::Exception& ex)
     {
         std::cerr << "Caught except::exception: " << ex.getMessage()
                   << std::endl;
-		TEST_ASSERT(test_assert_false);
+        TEST_ASSERT(false);
     }
     catch (...)
     {
         std::cerr << "Caught unknown exception\n";
-		TEST_ASSERT(test_assert_false);
+        TEST_ASSERT(false);
     }
 }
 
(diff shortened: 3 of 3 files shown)


build/build.py
```diff
@@ -969,9 +969,10 @@ def configureCompilerOptions(self):
         # specifications properly.
         # Skipping warning 4512 about being unable to generate an assignment
         # operator (since we often do this intentionally).
+	# warning C4100: '...' : unreferenced formal parameter
         # For warnings, use /W4 because /Wall
         # gives us tons of warnings in the VS headers themselves
-        warningFlags = '/W4 /wd4290 /wd4512'
+        warningFlags = '/W4 /wd4290 /wd4512 /wd4100'
```
asylvest: The only minor hesitation I have is that if we have code in a header file which has unreferenced formal parameters (only something inlined or templated), and we don't remove the unreferenced parameter names because we don't see this warning, and then someone else is using our stuff in their own build system where they don't put in this warning flag, they will suffer through extra warnings.  That being said, I think it's worth the tradeoff vs. practically speaking going through all the code and updating for this, so I think this is good.

modules/c++/io/tests/serializeTest1.cpp
```diff
@@ -73,7 +73,9 @@ class A : public Serializable
     {
 	std::string toFill;
 	sys::byte b[1];
-	while (true)
+	// Work-around warning C4127: conditional expression is constant;
+	// see https://msdn.microsoft.com/en-us/library/6t66728h.aspx and http://stackoverflow.com/questions/1946445/c-c-how-to-use-the-do-while0-construct-without-compiler-warnings-like-c412
+	for (;;) // while (true)
```
asylvest: Rather than this, is there a `/wd4127` we can put in `build.py`?  `while (true)` is more instantly obvious to people what's going on I think.

modules/c++/math.linear/include/math/linear/MatrixMxN.h
```diff
@@ -39,7 +39,17 @@ template<typename _T> bool equals(const _T& e1, const _T& e2)
 
 template<typename _T> inline bool equals(const _T& e1, const _T& e2, _T eps)
 {
+#ifdef _MSC_VER // Visual Studio
```
asylvest: Similarly, to avoid needing to sprinkle this through all our code, can we just put in `/wd4127`?

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -100,7 +100,7 @@ sys::Pid_T sys::OSWin32::getProcessId() const
 
 void sys::OSWin32::removeFile(const std::string& pathname) const
 {
-    if (DeleteFile(pathname.c_str()) != true)
+    if (DeleteFile(pathname.c_str()) != TRUE)
```
asylvest: I am assuming that windows.h guarantees there's something called `TRUE` set to non-zero?

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -113,7 +113,7 @@ void sys::OSWin32::removeFile(const std::string& pathname) const
 
 void sys::OSWin32::removeDirectory(const std::string& pathname) const
 {
-    if (RemoveDirectory(pathname.c_str()) != true)
+	if (RemoveDirectory(pathname.c_str()) != TRUE)
```
asylvest: Looks like whitespace got added here

modules/c++/sys/unittests/test_symlink.cpp
```diff
@@ -67,18 +69,18 @@ TEST_CASE(testCreateSymlink)
     catch (const std::exception& ex)
     {
         std::cerr << "Caught std::exception: " << ex.what() << std::endl;
-        TEST_ASSERT(false);
```
asylvest: If this change is just to get rid of the compiler warning, hoping we can take care of that in build.py

build/build.py
JDanielSmith: I've always used the "unreferenced parameter" warning as a not-so-subtle hint saying "maybe you should GET RID of this parameter...after all, you're not using it."  Obviously, that doesn't work for a situation like overriding a virtual function.

My personal preference would be to make code changes to remove the warning; but that would seem like a (very) low priority item.  And in the meantime, this warning (and others) might hide something more important.

asylvest: Agreed on both accounts.  Spawned #134 for this to work on in some mythical day in the future when all our other problems have been resolved.

modules/c++/io/tests/serializeTest1.cpp
JDanielSmith: /wd4127 can find other less obvious situations than "while (true)".  The warning itself can be somewhat interesting because in a situation like "if (... something that is true at compile time ...)" the "else" (if any) will never be executed, which might indicate dead code or a programming error.

Yes, I would also prefer "while (true)" which is why I left it in as a comment.

modules/c++/math.linear/include/math/linear/MatrixMxN.h
JDanielSmith: I would rather fix this as described here: http://stackoverflow.com/questions/25573996/c4127-conditional-expression-is-constant; i.e., use template magic so that everything is evaluated at compile-time--which from the warning we already know will work.

But that seemed like too much code for just this one place, and it looks like std::conditional<> in C++14 might provide a better fix: http://en.cppreference.com/w/cpp/types/conditional

If a conditional expression really is constant at compile time, why is it a conditional at all?  Nothing is going to change during program execution, and the condition won't vary.

build/build.py
JDanielSmith: "... some mythical day in the future when all our other problems have been resolved." :-)

modules/c++/sys/unittests/test_symlink.cpp
JDanielSmith: Adding /wd4127 to the command-line makes it harder to get to the day when the code is free of the "conditional expression is constant" warning, and no new code can be added which generates that warning because "warnings as errors" is enabled.

Of course, it's also perfectly reasonable to say "we'll never, ever care about this diagnostic" and always turn it off.  I don't like doing that because I figure the compiler writer is smarter than I am and the warning is there for a reason.  Maybe not a very big or important reason (right now), but something that I should perhaps think about some day.

modules/c++/sys/source/OSWin32.cpp
JDanielSmith: Yes, although the actual file is <minwindef.h>, which gets #included from <windows.h>.

This code might better be tweaked to be "if (!DeleteFile(...))" or "if (DeleteFile(...) == FALSE)" as the docs say "If the function succeeds, the return value is nonzero. If the function fails, the return value is zero (0)."

modules/c++/io/tests/serializeTest1.cpp
asylvest: Yeah, it is a tradeoff.  The other thing is that a good chunk of our compilation is on Linux (where warnings are much more sparse because otherwise I would go crazy) and even with `-Wall` on gcc doesn't emit this one, so arguably we've been getting by fine without it.  I wish there were more fine-grained warnings you could turn on/off such that you could tell the compiler that the really obvious stuff like `while (true)` was done with intention and so doesn't need to be a warning, though the compiler would have to be pretty smart to distinguish that I suppose.

To me, the possibility of someday making a logic error that leads to dead code is worth the risk to avoid needing to explain to every new hire that they need to write `for (;;)`, so I'd vote to add this to build.py.

modules/c++/math.linear/include/math/linear/MatrixMxN.h
asylvest: Given that the compiler knows that it's a constant at compile time, it's almost certainly going to optimize it out when compiling for release, so there's not going to be a performance hit.  But I do agree that it's unsatisfying the way it's structured now.

At first I liked the `condition` class on Stack Overflow, but thinking about it a bit more, aside from the advantage of skirting around the compiler warning, it's still the same thing... you have an `if` statement with a condition that the compiler can evaluate at compile time based on which template specialization it picks and then presumably inlines it away.

I've done some stuff with boost back in the day where they had a `true_type` and `false_type`.  Then you made a function overloading where the last parameter was either this true or false type, they had a similar condition type object which presumably again used template specialization to return the correct type... but then here there truly wasn't an `if` statement; it was the compiler choosing the correct function overloading which of course happened at compile time.  Something notionally like:

```
template <typename T>
bool equals(const T& e1, const T& e2, T eps, boost::false_type)
{
    return std::abs(e1 - e2) < eps;
}

template <typename T>
bool equals(const T& e1, const T& e2, T eps, boost::true_type)
{
    if (e1 == std::numeric_limits<T>::infinity() ||
        e1 == -std::numeric_limits<T>::infinity())
    {
        return e1 == e2;
    }

    return equals(e1, e2, eps, boost::false_type);
}

template <typename T>
bool equals(const T& e1, const T& e2, T eps)
{
    return equals(e1, e2, eps, boost::conditional(std::numeric_limits<T>::has_infinity));
}
```

We're (unfortunately) not heavily using boost right now, though I'm open to changing that when we have a compelling reason... it's mostly due to momentum that we hadn't used it in the past for these projects.  But here, I think we can pretty easily accomplish what we want without boost by still using template specialization.  I didn't actually compile this but the syntax should be close:

```
template <bool b>
struct Equals
{
    template <typename T>
    static bool compare()(const T& e1, const T& e2, T eps)
    {
        return std::abs(e1 - e2) < eps;
    }
};

template <>
struct Equals<true>
{
    template <typename T>
    static bool compare()(const T& e1, const T& e2, T eps)
    {
        if (e1 == std::numeric_limits<T>::infinity() ||
            e1 == -std::numeric_limits<T>::infinity())
        {
            return e1 == e2;
        }
        
        return Equals<false>::compare(e1, e2, eps);
    }
};

template <typename T>
bool equals(const T& e1, const T& e2, T eps)
{
    return Equals<std::numeric_limits<T>::has_infinity>::compare(e1, e2, eps);
}
```

Obviously someone needs to understand more C++ to see what's going on here, and it's increased the number of lines of code, but I think this is legible if you want to put something like this together.  Make sure to add unit tests if there aren't some already! :o)

modules/c++/sys/source/OSWin32.cpp
asylvest: Yeah I recall a lot of this code being guilty of that.  I wouldn't bother spending the time cleaning up all of the existing functions, but for ones that you've already touched here due to the compiler warning, yeah, go ahead and change to be `if (DeleteFile(...) == FALSE)`

modules/c++/math.linear/include/math/linear/MatrixMxN.h
JDanielSmith: The boost code is about the same as one of the solutions from SO.  I tried it out (and kinda liked it--even if it is rather heavy on C++ templates), but it seemed like an awful lot of (isolated) code to add for this one (rather spurious) warning.

Besides, this will go away (too) if we add /wd4127 to the command-line as you're suggesting above.

modules/c++/io/tests/serializeTest1.cpp
JDanielSmith: See https://blogs.msdn.microsoft.com/vcblog/2016/06/03/c-core-guidelines-checkers-are-now-in-a-single-nuget-package/ (links to http://reviews.llvm.org/diffusion/L/browse/clang-tools-extra/trunk/clang-tidy/cppcoreguidelines/) for one way to get (a lot?) more warnings on Linux.

This (http://stackoverflow.com/a/12247461/8877) SO answer indicates that GCC also has a lot more warnings available than those from -Wall.

JDanielSmith:
Far too many other changes have been made ... no use trying to do anything with this request anymore.

PR #137, opened 2016-06-14
If you specify a build target (python waf install target='x'), waf wi…

JonathanMeans:
…ll only install that target. Update to ensure entire project gets installed when targets are included


asylvest:
Not needed - we changed how Jenkins was set up

PR #138, opened 2016-06-29
Add support for using C++11's regex rather than PCRE.  Also upgrade PCRE version to 10.22

asylvest:
@mwysockiMDA FYI in the future when you are done with a branch, create a pull request like this one


mwysockiMDA:
Okay, but I just noticed it fails to compile right now without --enable-cpp11, so it actually shouldn't be merged quite yet.

asylvest:
No prob - this is the mechanism I'll use to put annoying review comments inline with the code :o)

mwysockiMDA:
OK, gotcha

modules/c++/re/include/re/PCRE.h
```diff
@@ -29,10 +29,18 @@
 #define USE_PCRE
 #endif
 
+#include <sys/sys_config.h>
```
asylvest: So... the `re` module sounds like it's a generic regex module, but then it names the actual class that you interact with directly after the library that it uses for implementation.  This gets extra confusing now that we have an STL implementation.  How about renaming PCRE.h --> Regex.h (and rename class to `Regex`), PCRE.cpp --> RegexPCRE.cpp, PCRE_STL.cpp --> RegexSTL.cpp, PCREException --> RegexException.  Then rename unit tests and calling code appropriately.  Obviously the downside of this is this will ripple through all the calling code but I think it'll be named more correctly at that point.

asylvest:
Overall, I think how this would ideally work is that if you configured with `--enable-cpp11`, it would not configure or build modules/drivers/pcre (maybe there should be a configure option to force it to pick PCRE over C++11).  So, this will have a ripple effect on how some wscripts and ifdefs in the code are set up

asylvest:
Along those lines, the...

```
#if !defined(USE_PCRE)
// Why would this be included if it wasn't intended to be used?
#define USE_PCRE
#endif
```

that's been in PCRE.h should be updated

modules/c++/re/include/re/PCRE.h
```diff
@@ -171,10 +187,22 @@ class PCRE
...
+     *  \return  The modified string
+     */
+    std::string replace_dot(const std::string& str) const;
+
+    std::regex mRegex;
+#else
     //! The pcre object
     pcre *      mPCRE;
     //! The output/offset vector
     int         mOvector[OVECCOUNT];
```
asylvest: I know this predates your changes but any reason why the `OVECCOUNT` constant that's at the top of this file currently can't move to be a private constant?

modules/c++/re/include/re/PCRE.h
```diff
@@ -171,10 +187,22 @@ class PCRE
 
 protected:
```
asylvest: I know this predates your changes but any reason this can't be private rather than protected?

mwysockiMDA:
Yeah, something else I found is that --enable-cpp11 fails with VS2013 (it tests for constexpr support). So that means if we want to use std::regex in Windows, we need to break the C++11 test into multiple parts with multiple flags.

asylvest:
It should work with VS 2015 I believe... if you don't have 2015 Express on your machine, put in a ticket to get that.  It's ok to assume that if someone wants C++11 on Windows, they will use VS 2015.

modules/c++/re/include/re/PCRE.h
```diff
@@ -171,10 +187,22 @@ class PCRE
...
     std::string mPattern;
+
+#ifdef __CODA_CPP11
+    /*!
+     *  Replace non-escaped "." with "[\s\S]" to get PCRE_DOTALL newline behavior
+     *  \param str  The string to modify
+     *  \return  The modified string
+     */
+    std::string replace_dot(const std::string& str) const;
```
asylvest: Rename to `replaceDot()`

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+re::PCRE::~PCRE()
+{
+    destroy();
+}
+
+re::PCRE::PCRE(const re::PCRE& rhs)
+{
+    mPattern = rhs.mPattern;
```
asylvest: Does this and the assignment operator need the `if (!mPattern.empty())` checks or does it not matter?

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+                            int flags)
+{
+    mPattern = (flags==PCRE_DOTALL) ? replace_dot(pattern) : pattern;
+    try 
+    {
+        mRegex = std::regex(mPattern);
+    }
+    catch (const std::regex_error& e)
+    {
+        throw PCREException(Ctxt(FmtX("PCRE std::regex constructor error: %s",
```
asylvest: A lot of our older code uses `FmtX()` but actually try to avoid it since internally it's using a fixed-size char buffer which is just begging to overflow.  So just change to something like...

`throw PCREException(Ctxt("PCRE std::regex constructor error: " + std::string(e.what())))`

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+#include "re/PCRE.h" // this has to come before the #ifdef checks below
+
+// we use this file if we're not using actual PCRE itself
+#if defined(USE_PCRE) && defined(__CODA_CPP11)
+
+re::PCRE::PCRE(const std::string& pattern, int flags) :
+    mPattern(pattern)
+{
+    if (!mPattern.empty())
```
asylvest: Style nit - always use curly braces... even for one-liner conditionals like this (motivation is to not add a second line in the future and not notice that second line isn't actually inside the `if`)

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+bool re::PCRE::match(const std::string& str,
+                     PCREMatch & matchObject,
+                     int )
+{
+    std::smatch matches;
+    bool result = std::regex_search(str, matches, mRegex);
+
+    // copy resulting substrings into matchObject
+    matchObject.resize( matches.size() );
```
asylvest: Don't need extra spaces around parens... so just `matchObject.resize(matches.size());`

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    // This causes a crash for some reason
+    //std::copy(matches.begin(), matches.end(), matchObject.begin());
+
+    for(size_t i=0; i<matches.size(); ++i) {
```
asylvest: * Avoid single character variable names (change to `ii`)
* Put curly brace on its own line

asylvest: Ahh and there should be some more spaces in there...

`for (size_t ii = 0; ii < matches.size(); ++ii)`

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    std::smatch matches;
+
+    // search the string starting at index "startIndex"
+    bool result = std::regex_search(matchString.begin()+startIndex, 
+                                    matchString.end(), matches, mRegex);
+    
+    // if successful, return the substring matching the regex,
+    // otherwise return empty string
+    if (result && matches.size() > 0)
```
asylvest: Does `smatch` have a `.empty()` that you can change the second half to `&& !matches.empty()`?

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        return matches[0].str();
+    }
+    else
+    {
+        return std::string("");
```
asylvest: Can just do `return ""`

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+}
+
+void re::PCRE::searchAll(const std::string& matchString,
+                         PCREMatch& v)
+{
+    // this iterates mRegex over the input string, returning an
+    // iterator to the match objects
+    auto words_begin =
```
asylvest: camelCase for these variable names

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+    for (std::sregex_iterator match_iter = words_begin; 
+         match_iter != words_end;++match_iter)
+    {
+        std::string match_str = match_iter->str(); 
+        v.push_back(match_str);
+    }
+}
+
+void re::PCRE::split(const std::string& str,
+                     std::vector< std::string > & v)
```
asylvest: No extra spaces around `<>`

modules/c++/re/include/re/PCRE.h
mwysockiMDA: Done.

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    int idx = 0;
```
asylvest: Can be `size_t` rather than `int` right?

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+    std::smatch match;
+    while (std::regex_search(str.begin()+idx, str.end(), match, mRegex, flags))
+    {
+        v.push_back( str.substr(idx, match.position()) );
+        idx += (match.position() + match.length());
+
+        // not sure this will ever be needed for a split() operation,
+        // but we'll be safe (matches after first match will not match
+        // beginning-of-line))
+        if(flags == std::regex_constants::match_default)
```
asylvest: Add a space between `if` and first paren

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+        v.push_back(str.substr(idx));
+    }
+}
+
+std::string re::PCRE::sub(const std::string& str,
+                          const std::string& repl)
+{
+    std::string toReplace = str;
+
+    int idx = 0;
```
asylvest: `idx` can be `size_t`?

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+    auto flags = std::regex_constants::match_default;
+    std::smatch match;
+    while (std::regex_search(toReplace.cbegin()+idx, toReplace.cend(), match, mRegex, flags))
```
asylvest: 80 character line limit

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+            flags = std::regex_constants::match_not_bol;
+    }
+
+    return toReplace;
+}
+
+std::string re::PCRE::escape(const std::string& str) const
+{
+    std::string r;
+    for (unsigned int i = 0; i < str.length(); i++)
```
asylvest: `size_t` rather than `unsigned int`

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        if (!isalpha(str.at(i)) && !isspace(str.at(i)))
```
asylvest: Since you know you're in-bounds, just use `str[i]` rather than `str.at(i)`

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+        {
+            r += '\\';
+        }
+        r += str.at(i);
```
asylvest: Ditto

modules/c++/re/source/PCRE_STL.cpp
```diff
@@ -0,0 +1,219 @@
...
+        compile(mPattern, flags);
+}
+
+void re::PCRE::destroy()
+{
+}
+
+re::PCRE::~PCRE()
+{
+    destroy();
```
asylvest: Since you know in the STL implementation that `destroy()` is a noop, I think it's fair to remove it from the destructor and assignment operator

asylvest:
Does any calling code ever send flags into the `match()`, `matches()`, or `search()` methods?  If not, for safety I think we should remove this from the API so that no one's throwing in some crazy PCRE flag we're not aware of in some other codebase out there and we start ignoring it in the STL implementation.

modules/c++/re/source/PCRE_STL.cpp
mwysockiMDA: I made the change, but doesn't that just implicitly call the std::string constructor?

asylvest: Yeah the same code will get generated... so it's just a style / OCD nit

modules/c++/re/source/PCRE_STL.cpp
mwysockiMDA: It's operator= on two std::strings, so it shouldn't do anything if rhs is empty.

modules/c++/re/include/re/Regex.h
```diff
@@ -0,0 +1,203 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public 
+ * License along with this program; If not, 
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+#ifndef __RE_PCRE_H__
```
asylvest: Change include guard to match filename

modules/c++/re/include/re/Regex.h
```diff
@@ -0,0 +1,203 @@
...
+#include <pcre.h>
+#include <pcreposix.h>
+#endif
+
+#include <string>
+#include <vector>
+
+/*!
+ *  \file Regex.h
+ *  \brief C++ wrapper for the PCRE library
```
asylvest: Update description

modules/c++/re/include/re/Regex.h
```diff
@@ -0,0 +1,203 @@
...
+ */
+
+namespace re
+{
+
+    typedef std::vector<std::string> RegexMatch;
+    /*!
+     *  \class Regex
+     *  \brief C++ wrapper object for the PCRE library,
```
asylvest: Update description

modules/c++/re/include/re/Regex.h
```diff
@@ -0,0 +1,203 @@
...
+
+        bool matches(const std::string& str) const;
+
+        /*!
+         *  Search the matchString
+         *  See the pcre man pages (pcre_exec) for further
+         *  info on the optional flags.
+         *  \param matchString The string to try and match
+         *  \param startIndex Starting where?
+         *  \param flags  Any flags
```
asylvest: Remove description

modules/c++/re/include/re/Regex.h
```diff
@@ -0,0 +1,203 @@
...
+         *  \return  The escaped string
+         */
+        std::string escape(const std::string& str) const;
+
+    private:
+        std::string mPattern;
+
+#ifdef __CODA_CPP11
+        /*!
+         *  Replace non-escaped "." with "[\s\S]" to get Regex_DOTALL newline behavior
```
asylvest: Is `Regex_DOTALL` supposed to be `PCRE_DOTALL` from a find/replace?

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -20,39 +20,41 @@
  *
  */
 
-#if defined(USE_PCRE)
+#include "re/Regex.h" // this has to come before the #ifdef checks below
```
asylvest: Doesn't matter that much I guess but why is it the case that this include has to come before the `#ifdef` check?

modules/c++/re/include/re/Regex.h
```diff
@@ -0,0 +1,203 @@
...
+#else
+        // Internal function for passing flags to pcre_exec()
+        std::string search(const std::string& matchString,
+                           int startIndex, int flag);
+
+        // Size of the output vector, must be a multiple of 3
+        // The output vector is filled up to 2/3 (666) full for matches
+        // so the maximum number of substrings is 333 (333 start
+        // offsets and 333 end offsets)
+        const int OVECCOUNT;
```
asylvest: If this is a member variable now (didn't it used to be a static constant?), then it should change over to a normal member variable name like `mOvectorCount`.  But since `mOvector` is an STL vector, can't you just use `mOvector.size()` throughout and get rid of `OVECCOUNT` altogether?

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -20,39 +20,41 @@
...
 
-re::PCRE::~PCRE()
+re::Regex::~Regex()
 {
     destroy();
 }
 
-re::PCRE::PCRE(const re::PCRE& rhs)
+re::Regex::Regex(const re::Regex& rhs) :
+    mPattern(rhs.mPattern), OVECCOUNT(999), mPCRE(NULL), mOvector(OVECCOUNT)
```
asylvest: Don't want to get out of sync with having 999 in two spots if we can help it.  Should be able to set from `rhs.OVECCOUNT`) or if you ditch the variable altogether, size `mOvector` based on `rhs`'s size.

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -87,43 +89,44 @@ re::PCRE& re::PCRE::compile(const std::string& pattern,
 
     if (mPCRE == NULL)
     {
-        throw PCREException(Ctxt(FmtX("PCRE compile error at offset %d: %s",
-                                      erroffset, errorptr)));
+        std::stringstream ss;
+        ss << "Regex compile error at offset "
```
asylvest: "Regex" should be "PCRE" shouldn't it?

modules/c++/re/source/RegexSTL.cpp
```diff
@@ -0,0 +1,292 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "re/Regex.h" // this has to come before the #ifdef checks below
```
asylvest: Again, why does this include have to come first?

modules/c++/re/source/RegexSTL.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+// we use this file if we're not using actual PCRE itself
+#if defined(__CODA_CPP11)
```
asylvest: Can simplify to just `#ifdef __CODA_CPP11`

modules/c++/re/source/RegexSTL.cpp
```diff
@@ -0,0 +1,292 @@
...
+        compile(mPattern);
+    }
+
+    return *this;
+}
+
+re::Regex& re::Regex::compile(const std::string& pattern)
+{
+    mPattern = replaceDot(pattern);
+    try
```
asylvest: Rather than re-catching, just let the STL exception go (any caller that cares should be at some level in their application catching both CODA exceptions and STL exceptions).  Motivation is just for code simplicity / maintenance reasons - since the STL already behaves like we want, in general I try to avoid the extra couple lines for the re-catch unless we need to provide a specialized exception message.  So then update Regex.h's description to just say it throws in general rather than a specific exception that it throws.

modules/c++/re/source/RegexSTL.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+void re::Regex::searchAll(const std::string& matchString,
+                          RegexMatch& v)
+{
+    std::smatch match;
+    size_t startIndex = 0;
+    bool matchBeginning = true;
+
+    // search the string starting at index "startIndex"
+    while( searchWithContext(matchString.begin()+startIndex,
```
asylvest: Change whitespace to `while (`

modules/c++/re/source/RegexSTL.cpp
```diff
@@ -0,0 +1,292 @@
...
+    // Because VS2015 and gcc handle ^ and $ differently, we'll throw
+    // exceptions if they're in the middle of the pattern somewhere
+
+    // Look for ^ in the middle, but ignore \^ and [^
+    if (std::regex_search(mPattern, tmpmatch, 
+                          std::regex(R"lit([\s\S]*([^\[\\]|[^\\](\\\\)+)\^)lit"),
+                          std::regex_constants::match_continuous))
+    {
+        std::string msg(
+            "RegexSTL: '^' in mid-string is not handled the same by gcc and VS2015!");
```
asylvest: Don't need "RegexSTL" prefix since the `Ctxt()` macro will provide that

modules/c++/re/source/RegexSTL.cpp
```diff
@@ -0,0 +1,292 @@
...
+        msg += " So we don't allow it :(";
+        throw RegexException(Ctxt(msg));
+    }
+
+    // Look for $ in the middle, but ignore \$
+    if (std::regex_match(mPattern, tmpmatch,
+                         std::regex(R"lit(^([\s\S]*[^\\](\\\\)*)?\$[\s\S]+$)lit")))
+    {
+        std::string msg(
+            "RegexSTL: '$' in mid-string is not handled the same by gcc and VS2015!");
```
asylvest: Ditto

modules/c++/re/source/RegexSTL.cpp
```diff
@@ -0,0 +1,292 @@
...
+        else
+        {
+            flags |= std::regex_constants::match_continuous;
+            b = std::regex_search(inputIterBegin, inputIterEnd,
+                                  match, mRegex, flags);
+        }
+    }
+    else if (!mPattern.empty() && mPattern.back() == '$')
+    {
+        std::string msg("RegexSTL: trailing '$' will not be handled correctly!");
```
asylvest: Ditto

modules/c++/re/unittests/test_regex.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+TEST_CASE(testDotAllFlag)
+{
+    // This should match both the "3.3" and "4\n2"
+    re::RegexMatch matches1;
+    re::Regex rx1("\\d.\\d");
+    rx1.searchAll("3.3 4\n2", matches1);
+    TEST_ASSERT_EQ(matches1.size(), 2);
+
+    // This should only match the "3.3"
```
asylvest: Is this supposed to still be commented out?

modules/c++/re/unittests/test_regex.cpp
```diff
@@ -0,0 +1,312 @@
...
+{
+    TEST_CHECK( testCompile);
+    TEST_CHECK( testMatches);
+    TEST_CHECK( testSearch);
+    TEST_CHECK( testSearchAll);
+    TEST_CHECK( testDotAllFlag);
+    TEST_CHECK( testMultilineBehavior);
+    TEST_CHECK( testSub);
+    TEST_CHECK( testSplit);
+    TEST_CHECK( testHttpResponse);
```
asylvest: Add a `return 0` at the end

modules/c++/re/include/re/Regex.h
mwysockiMDA: I didn't want to modify it in case it's a PCRE limitation.  I've changed it to a static const private member so it is shared across instances and only initialized once.

modules/c++/re/source/RegexPCRE.cpp
mwysockiMDA: Changed to static const int private member.

modules/c++/re/source/RegexPCRE.cpp
mwysockiMDA: Ah, that was because it defined USE_PCRE which is gone now. I rearranged it.

modules/c++/re/unittests/test_regex.cpp
mwysockiMDA: Yeah, that test was from before I removed the public DOTALL flag.  I removed the commented-out code.

modules/c++/re/source/RegexPCRE.cpp
mwysockiMDA: Oops, nevermind. That was actually because the __CODA_CPP11 define is #included from sys_config.h in Regex.h, so that has to be #included before the #ifdef.

modules/c++/re/source/RegexPCRE.cpp
asylvest: OK sorry one more name change here... if it's a static constant, then it should be all caps and with underscores between the words to make it really obvious that it's static (and the `m` no longer applies).  So, change to `OVECTOR_COUNT`.

asylvest:
Looks good - just that one more variable to rename.  Also, are RegexTest2 and RegexTest3 still testing unique things or are they all covered by the test_regex unittest such that one or both of them could be removed?

mwysockiMDA:
The two tests looked almost identical, except that RegexTest3 was structured in a class like it might have been copied from a real object, so I incorporated RegexTest3 into the unit tests.  I found the test files useful because they print so much more than the unit tests, but they can go away if needed.

asylvest:
OK no prob - you can leave the original test programs around

modules/c++/re/unittests/test_regex.cpp
```diff
@@ -0,0 +1,313 @@
...
+TEST_CASE(testCompile)
+{
+    re::Regex rx;
+    // test that an invalid regexp throws an exception
+    try
+    {
+        rx.compile("^(");
+        TEST_FAIL("Should have thrown an exception for this invalid regular expression!");
+    }
+    catch (...)
```
asylvest: Ahh this is lame of TestCase.h.  Can you add a new macro to it `TEST_THROWS` that does a `catch (...)` and then use that here?

modules/c++/re/tests/benchmark_regex.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+int main(int argc, char** argv)
+{
+    if (argc < 3)
+    {
+        std::cerr << "Usage: ./bm_regex inputFile numIterations" << std::endl;
+        return 1;
+    }
+
+    uint64_t numIterations = std::stoi(argv[2]);
```
asylvest: We live in a sad world where C++11 is no guarantee.  Use `str::toType()` rather than `std::stoi()`.  Also, to protect against really lame old versions of Visual Studio that don't have `stdint.h`, rather than using `uint64_t` directly, use `sys::Uint64_T`.

modules/c++/re/tests/benchmark_regex.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+// First just benchmark the Regex creation time (including compile)
+double BM_RegexCreation(uint64_t numIterations)
+{
+    sys::RealTimeStopWatch sw;
+
+    sw.start();
+    for (uint64_t ii = 0; ii < numIterations; ++ii)
+    {
+        re::Regex regex("beam(Id|String)");
```
asylvest: Hmm, I guess we should take this string in on the command line too so that we can tailor it to what we're parsing

modules/c++/re/tests/benchmark_regex.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    double elapsedTimeMS = sw.stop();
+
+    return elapsedTimeMS / numIterations;
+}
+
+
+int main(int argc, char** argv)
+{
+    if (argc < 3)
```
asylvest: Surround the whole thing in a try/catch block similar to what other test programs / benchmarks do

modules/c++/re/tests/benchmark_regex.cpp
```diff
@@ -0,0 +1,96 @@
...
+    {
+        std::cerr << "Usage: ./bm_regex inputFile numIterations" << std::endl;
```
asylvest: Use `sys::Path::basename(argv[0])` rather than hardcoding in the application name

modules/c++/re/tests/benchmark_regex.cpp
```diff
@@ -0,0 +1,96 @@
...
+    // Now run the benchmarks
+    double swtime0 = BM_RegexCreation(numIterations);
+    double swtime1 = BM_RegexMatch(numIterations, fileString);
+
+    // Convert ms to ns
+    swtime0 *= 1.e6;
+    swtime1 *= 1.e6;
+
+    // Pretty-print our results
+    char outbuff[10000];
```
asylvest: Is it a huge pain to use `setw` and friends here rather than `sprintf()`?

modules/c++/re/tests/benchmark_regex.cpp
```diff
@@ -0,0 +1,96 @@
...
+    {
+        std::cerr << "Error opening text file!" << std::endl;
+        return 2;
+    }
+    
+    size_t size = bigFin.tellg();
+    std::string fileString(size, '\0');
+    
+    bigFin.seekg(0);
+    bigFin.read(&fileString[0], size);
```
asylvest: It's not guaranteed that this is safe to write into is it (even though presumably the underlying implementation is a char array)?  I've always used a `std::vector<char>` for this sort of thing...

mwysockiMDA: Apparently this was changed in C++11 (which outlawed Copy-on-Write implementations of std::string like gcc's). I'll switch to `vector<char>`.

modules/c++/re/tests/benchmark_regex.cpp
mwysockiMDA: No, but I'll die a little inside.  `boost::format()` would work great here (wink wink).

modules/c++/re/tests/benchmark_regex.cpp
mwysockiMDA: Made the change, but out of curiosity, what's the reasoning behind this?  For a test program I'd actually rather have an uncaught exception so I can do a backtrace in gdb.

modules/c++/re/tests/benchmark_regex.cpp
asylvest: Ha!

asylvest:
See #160 for some additional discussions.

Long story short:
- There is a generic `re::Regex` class which generalizes our regex support.  API looks very similar to what it did previously (but the old class was named `PCRE`).
- You can now configure with `--enable-std-regex` to use the STL regex library for the underlying implementation.  Doing this, PCRE is no longer required or built.  Unfortunately, performance is 4-5x slower than PCRE.  This doesn't much matter for one-off regex matches that aren't performance-critical, but for cases where there are hundreds of regexes, the difference is noticeable.
- The underlying PCRE version was upgraded from 5.0 to 10.22.  The underlying PCRE API had changed; one thing that did get simpler in the wscript is we no longer have to build the separate `dftables` application... PCRE ships with reasonable tables that work fine for ASCII.  This has been tested on Linux and Windows.  On Linux it also works properly with `-shared`.  On Windows this can't be tested currently due to some kind of waf problem with our `-shared` support in general (#165).

asylvest:
I believe all work on this is complete.  Waiting on @msciaramitaro to have a chance to test in back on a larger application before merging into master.

asylvest:
@msciaramitaro verified this is working properly in our other applications so this is (finally!) good to go.

mwysockiMDA:
:joy:

PR #145, opened 2016-07-11
Extended python coda_types to support pickling.

tolsonMDA:
By default, swig-wrapped objects do not play nicely with pickle (because the python layer doesn’t necessarily know how to serialize the c++ data), so this revision adds python extensions for the templated types defined in types.i.

### DIFF
--- a/modules/python/types/source/types.i
+++ b/modules/python/types/source/types.i
@@ -40,6 +40,52 @@
 %include "types/RowCol.h"
 %include "types/RgAz.h"
 
+// Pickle utilities
+%pythoncode
+%{
+    import cPickle as pickle
+%}
+
+%extend types::RowCol
+{
+%pythoncode
+%{
+    def __getstate__(self):
+        return (self.row, self.col)
+
+    def __setstate__(self, state):
+        self.__init__(state[0], state[1])
+%}
+}
+
+%extend types::RgAz
+{
+%pythoncode
+%{
+    def __getstate__(self):
+        return (self.rg, self.az)
+
+    def __setstate__(self, state):
+        self.__init__(state[0], state[1])
+%}
+}
+
+%extend std::vector
+{
+%pythoncode
+%{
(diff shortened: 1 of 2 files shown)


modules/python/types/source/types.i
```diff
@@ -47,3 +47,93 @@
...
 %template(VectorRowColInt) std::vector<types::RowCol<sys::SSize_T> >;
 %template(VectorSizeT) std::vector<size_t>;
+
+// Pickle utilities
+%pythoncode
+%{
+    import cPickle as pickle
+%}
+
+%extend types::RowCol<double>
```
asylvest: I assume there's no way to do this extension on the original templated class so that we don't have this repeated code for each type..?

asylvest: You could write a macro (see `mem.i` for an example).  Normally I detest macros in pure C++ code but give some thought to if you think it's warranted here if there aren't other options.

asylvest:
Is there a test you can add that shows that Pickling now works?

modules/python/types/source/types.i
tolsonMDA: Good question. I had looked into this originally and found that one can easily extend the c++ templates, but didn't come up with a way to wrap the python code. I'll put a little more thought into it and let you know what I learn.

tolsonMDA:
Sure I can make a test. I'll add it to a "test" directory in the types folder unless there's a better place?

asylvest:
Yep - call the directory `tests` rather than `test` but yeah that sounds good

tolsonMDA:
I uploaded a templated version of the extensions. I think this is about as general as it can be since the state details are specific to the object definitions. Do you agree?

I'll add tests next.

asylvest:
Nice - yeah this looks good.

tolsonMDA:
Should be all set.

asylvest:
Looks good - thanks

PR #146, opened 2016-07-13
Current path

JonathanMeans:


modules/c++/sys/source/OSUnix.cpp
```diff
@@ -346,6 +346,53 @@ void sys::OSUnix::getMemInfo(size_t &totalPhysMem, size_t &freePhysMem) const
...
+    possibleSymlinks.push_back(sys::Path::joinPaths(
+            sys::Path::delimiter()[0] + std::string("proc"),
+            sys::Path::joinPaths("self",
+            sys::Path::joinPaths("path", "a.out"))));
+
+    for (size_t ii = 0; ii < possibleSymlinks.size(); ++ii)
+    {
+        struct stat linkStatus;
+        const std::string path = possibleSymlinks[ii];
+        if (lstat(path.c_str(), &linkStatus) == 0)
```
asylvest: If `lstat()` fails, we should throw.  Probably cleanest to add a function in the unnamed namespace that looks something like...

```
void linkStat(const std::string& pathname, stat& linkStatus);
```

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -346,6 +346,53 @@ void sys::OSUnix::getMemInfo(size_t &totalPhysMem, size_t &freePhysMem) const
...
+
+    // Solaris
+    possibleSymlinks.push_back(sys::Path::joinPaths(
+            sys::Path::delimiter()[0] + std::string("proc"),
+            sys::Path::joinPaths("self",
+            sys::Path::joinPaths("path", "a.out"))));
+
+    for (size_t ii = 0; ii < possibleSymlinks.size(); ++ii)
+    {
+        struct stat linkStatus;
```
asylvest: I don't think you need the `struct` keyword here (unless the compiler was getting confused with the `stat()` function)

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -346,6 +346,53 @@ void sys::OSUnix::getMemInfo(size_t &totalPhysMem, size_t &freePhysMem) const
...
+        const std::string path = possibleSymlinks[ii];
+        if (lstat(path.c_str(), &linkStatus) == 0)
+        {
+            // Guessing a sane maximum based on Windows MAX_PATH
+            // If it's not enough, handle it through the base class
+            size_t bufferSize = 260 + 1;
+            char buffer[bufferSize];
```
asylvest: Can't put an array "on the stack" with a size set at runtime (even though here really this constant is known at compile time, it's not a static constant so it only sort of is).  gcc lets you get away with this (and likely does a dynamic allocation) but it's not portable.

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -346,6 +346,53 @@ void sys::OSUnix::getMemInfo(size_t &totalPhysMem, size_t &freePhysMem) const
...
+
+    for (size_t ii = 0; ii < possibleSymlinks.size(); ++ii)
+    {
+        struct stat linkStatus;
+        const std::string path = possibleSymlinks[ii];
+        if (lstat(path.c_str(), &linkStatus) == 0)
+        {
+            // Guessing a sane maximum based on Windows MAX_PATH
+            // If it's not enough, handle it through the base class
+            size_t bufferSize = 260 + 1;
```
asylvest: Use `PATH_MAX` instead

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -346,6 +346,53 @@ void sys::OSUnix::getMemInfo(size_t &totalPhysMem, size_t &freePhysMem) const
...
+            char buffer[bufferSize];
+
+            // readlink does not null-terminate anything
+            memset(buffer, 0, bufferSize);
+
+            if (readlink(path.c_str(), buffer, bufferSize) == bufferSize)
```
asylvest: If `readlink()` fails, I think you should throw instead... again, put a version in the unnamed namespace that throws for you

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -343,6 +343,25 @@ void sys::OSWin32::getMemInfo(size_t& totalPhysMem, size_t& freePhysMem) const
...
+
+    size_t bytesRead = GetModuleFileName(NULL, buffer, MAX_PATH + 1);
+
+    if (bytesRead == MAX_PATH + 1 || bytesRead == 0)
+    {
+        // Path may be up to 32,767 characters, so take a more manual
+        // approach instead of guess-and-checking our way up
+        return AbstractOS::getCurrentExecutable(argvPathname);
+    }
+    // Adding extra null-terminator because sometimes XP doesn't include one
```
asylvest: I don't see the null terminator getting added here..?

modules/c++/sys/source/OSUnix.cpp
JonathanMeans: It does get confused without struct. Everything else is fixed.

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -44,6 +44,26 @@
...
+{
+void linkStat(const std::string& pathname, struct stat& linkStatus)
+{
+    if (lstat(pathname.c_str(), &linkStatus) != 0)
+    {
+        throw except::Exception(Ctxt(strerror(errno)));
+    }
+}
+
+int readlinkOrError(const std::string& pathname, char* buffer)
```
asylvest: Return type should be void not int right?

JonathanMeans: Right.

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -44,6 +44,26 @@
...
+{
+void linkStat(const std::string& pathname, struct stat& linkStatus)
+{
+    if (lstat(pathname.c_str(), &linkStatus) != 0)
+    {
+        throw except::Exception(Ctxt(strerror(errno)));
+    }
+}
+
+void readlinkOrError(const std::string& pathname, char* buffer)
```
asylvest: Should return the result as a `std::string` here rather than taking in a `char*`.  That way the caller doesn't need to know any details about PATH_MAX or mess with any buffers.

modules/c++/sys/include/sys/AbstractOS.h
```diff
@@ -231,6 +231,12 @@ class AbstractOS
      */
     virtual void getMemInfo(size_t& totalPhysMem, size_t& freePhysMem) const = 0;
 
+    /*!
+     *  Get the absolute path to the current executable
```
asylvest: For all new functions added, use Doxygen comments.  Namely, put in each `\param` and a `\return`

modules/c++/sys/source/OSUnix.cpp
asylvest: Rename to `readLink()` for consistency

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -346,6 +373,39 @@ void sys::OSUnix::getMemInfo(size_t &totalPhysMem, size_t &freePhysMem) const
...
+    possibleSymlinks.push_back(sys::Path::joinPaths(
+            sys::Path::delimiter()[0] + std::string("proc"),
+            sys::Path::joinPaths("self",
+            sys::Path::joinPaths("path", "a.out"))));
+
+    for (size_t ii = 0; ii < possibleSymlinks.size(); ++ii)
+    {
+        struct stat linkStatus;
+        const std::string pathname = possibleSymlinks[ii];
+        linkStat(pathname, linkStatus);
```
asylvest: So actually... why is `linkStat()` called?  Nothing does anything with `linkStatus`.  Did a check or some other usage get removed in some of the refactoring?

JonathanMeans: Ah, so I was originally trying to use that to get the size of path. But then the stat.st_size was 0, which I figured had something to do with the target being the running program. Anyway, it’s pulled out now completely.

PR #147, opened 2016-07-19
Glibc backtrace

chvink:
This function can help with debugging in some situations. With some extra work I can include a solaris and windows implementation, too.

### DIFF
--- a/modules/c++/sys/source/Backtrace.cpp
+++ b/modules/c++/sys/source/Backtrace.cpp
@@ -10,18 +10,44 @@ static const size_t MAX_STACK_ENTRIES = 62;
 #include <execinfo.h>
 #include <cstdlib>
 
+namespace
+{
+
+//! RAII wrapper for stack symbols
+class BacktraceHelper
+{
+public:
+    BacktraceHelper(char** stackSymbols)
+        : mStackSymbols(stackSymbols)
+    {}
+
+    ~BacktraceHelper()
+    {
+        std::free(mStackSymbols);
+    }
+
+    std::string operator[](size_t idx)
+    {
+        return mStackSymbols[idx];
+    }
+private:
+    char** mStackSymbols;
+};
+
+}
+
 std::string sys::getBacktrace()
 {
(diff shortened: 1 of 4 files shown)


modules/c++/sys/source/Backtrace.cpp
```diff
@@ -0,0 +1,37 @@
+#include <sys/Backtrace.h>
+
+#include <sys/sys_config.h>
+#include <sstream>
+
+const size_t MAX_STACK_ENTRIES = 62;
```
asylvest: Add `static`

modules/c++/sys/source/Backtrace.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#ifdef HAVE_EXECINFO_H
+
+#include <execinfo.h>
+#include <cstdlib>
+
+std::string sys::getBacktrace()
+{
+    void * stackBuffer[MAX_STACK_ENTRIES];
```
asylvest: Spacing here and in other spots with pointers should be `void* stackBuffer`

modules/c++/sys/source/Backtrace.cpp
```diff
@@ -0,0 +1,37 @@
...
+    int currentStackSize = backtrace(stackBuffer, MAX_STACK_ENTRIES);
+    char ** stackSymbols = backtrace_symbols(stackBuffer, currentStackSize);
+
+    std::stringstream ss;
+    for (int ii = 0; ii < currentStackSize; ++ii)
+    {
+        ss << stackSymbols[ii] << std::endl;
+    }
+    std::free(stackSymbols);
```
asylvest: Use a RAII object so this can't leak

chvink:
The feedback should all be addressed.

asylvest:
So this is really cool.  I didn't know such a function existed.  Changes look good.  @chvink can you add an issue to support this on Windows too?  We don't need to do it now but this way we won't forget.

FYI to @clydestanfield @JDanielSmith @mwysockiMDA @msciaramitaro @jbruce2112 @mbojrab

PR #154, opened 2016-08-04
add negation operators to math.linear objects

mwysockiMDA:
Added negation operator to Matrix2D, MatrixMxN, Vector, and VectorN, as well as unit tests.  Now you can do:
`  Matrix2D B = -A;`

### DIFF
--- a/modules/c++/math.linear/include/math/linear/Matrix2D.h
+++ b/modules/c++/math.linear/include/math/linear/Matrix2D.h
@@ -1092,9 +1092,12 @@ public:
      */
     Matrix2D operator-() const
     {
-        Matrix2D mneg(*this);
-        std::transform(mneg.mRaw, mneg.mRaw+mneg.mMN, mneg.mRaw, std::negate<_T>());
-        return mneg;
+        Matrix2D neg(*this);
+        std::transform(neg.mRaw,
+                       neg.mRaw + neg.mMN,
+                       neg.mRaw,
+                       std::negate<_T>());
+        return neg;
     }
 
     /*!
--- a/modules/c++/math.linear/include/math/linear/Matrix2D.h
+++ b/modules/c++/math.linear/include/math/linear/Matrix2D.h
@@ -24,6 +24,7 @@
 
 #include <cmath>
 #include <algorithm>
+#include <functional>
 #include <import/sys.h>
 #include <mem/ScopedArray.h>
 #include <math/linear/MatrixMxN.h>
--- a/modules/c++/math.linear/include/math/linear/Matrix2D.h
+++ b/modules/c++/math.linear/include/math/linear/Matrix2D.h
@@ -1082,6 +1082,21 @@ public:
     }
 
     /*!
+     *  Negation operator;
+     *
+     *  \code
+           B = -A;
+     *  \endcode
+     *
+     */
+    Matrix2D operator-() const
+    {
+        Matrix2D mneg(*this);
+        std::transform(mneg.mRaw, mneg.mRaw+mneg.mMN, mneg.mRaw, std::negate<_T>());
+        return mneg;
+    }
+
+    /*!
      *  serialize out to a boost stream
      */
     template <class Archive>
--- a/modules/c++/math.linear/include/math/linear/MatrixMxN.h
+++ b/modules/c++/math.linear/include/math/linear/MatrixMxN.h
@@ -1189,8 +1189,26 @@ public:
         return multiply(mx);
     }
     
-
-
+    /*!
+     *  Negation operator;
+     *
+     *  \code
+           B = -A;
+     *  \endcode
+     *
+     */
+    Like_T operator-() const
+    {
+        Like_T mneg;
+        for (size_t i = 0; i < _MD; ++i)
+        {
+            for (size_t j = 0; j < _ND; ++j)
+            {
+                mneg.mRaw[i][j] = -mRaw[i][j];
+            }
+        }
+        return mneg;
+    }
 
 };
 
--- a/modules/c++/math.linear/include/math/linear/MatrixMxN.h
+++ b/modules/c++/math.linear/include/math/linear/MatrixMxN.h
@@ -1199,15 +1199,15 @@ public:
      */
     Like_T operator-() const
     {
-        Like_T mneg;
-        for (size_t i = 0; i < _MD; ++i)
+        Like_T neg;
+        for (size_t ii = 0; ii < _MD; ++ii)
         {
-            for (size_t j = 0; j < _ND; ++j)
+            for (size_t jj = 0; jj < _ND; ++jj)
             {
-                mneg.mRaw[i][j] = -mRaw[i][j];
+                neg.mRaw[ii][jj] = -mRaw[ii][jj];
             }
         }
-        return mneg;
+        return neg;
     }
 
 };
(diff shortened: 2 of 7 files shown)


modules/c++/math.linear/unittests/test_mx.cpp
```diff
@@ -81,6 +81,38 @@ TEST_CASE(testScaleMultiplyMxN)
     TEST_ASSERT_EQ(B, C);
 }
 
+TEST_CASE(testNegateMxN)
+{
+    Matrix3x3 A = identityMatrix<3, double>();
+    Matrix3x3 B = -A;
+    Matrix3x3 C = -B;
+    TEST_ASSERT_EQ(A, C);
```
asylvest: Technically if the negation operator did nothing, this unit test would still pass

modules/c++/math.linear/include/math/linear/Matrix2D.h
```diff
@@ -1082,6 +1082,21 @@ class Matrix2D
...
+     *
+     *  \code
+           B = -A;
+     *  \endcode
+     *
+     */
+    Matrix2D operator-() const
+    {
+        Matrix2D mneg(*this);
+        std::transform(mneg.mRaw, mneg.mRaw+mneg.mMN, mneg.mRaw, std::negate<_T>());
```
asylvest: Do you need to include <functional>?

asylvest: Bah.
`<functional>`

modules/c++/math.linear/unittests/test_mx.cpp
mwysockiMDA: Excellent point

modules/c++/math.linear/include/math/linear/Matrix2D.h
mwysockiMDA: Hm, compiler didn't complain (gcc or VS2015). I'll add it in case it's getting pulled in somewhere else.

asylvest: Yeah that's the motivation... to always make sure the header is self-contained.  In real life, `<functional>` is probably usually getting pulled in elsewhere but this way we're safe

modules/c++/math.linear/include/math/linear/Matrix2D.h
```diff
@@ -1082,6 +1083,21 @@ class Matrix2D
...
+     *  Negation operator;
+     *
+     *  \code
+           B = -A;
+     *  \endcode
+     *
+     */
+    Matrix2D operator-() const
+    {
+        Matrix2D mneg(*this);
```
asylvest: Another nit... member variables start with `m`.  No Hungarian notation aside from that.  So in general, avoid abbreviating with just an `m` as the first letter (but something like `max` is fine if it's obviously part of a word).  So here you'd use either `neg`, or if you wanted to denote that it's a matrix, `matNeg` or something like that.

modules/c++/math.linear/include/math/linear/Matrix2D.h
asylvest: 80 character line limit... also put spaces between operators.  So this should become something like...

```
std::transform(neg.mRaw,
               neg.mRaw + neg.mMN,
               neg.mRaw,
               std::negate<_T>());
```

or

```
std::transform(neg.mRaw, neg.mRaw + neg.mMN, neg.mRaw,
               std::negate<_T>());
```

Personally I like the first one best.

modules/c++/math.linear/include/math/linear/MatrixMxN.h
```diff
@@ -1189,8 +1189,26 @@ class MatrixMxN
...
+     *  Negation operator;
+     *
+     *  \code
+           B = -A;
+     *  \endcode
+     *
+     */
+    Like_T operator-() const
+    {
+        Like_T mneg;
```
asylvest: Ditto on variable renaming

modules/c++/math.linear/include/math/linear/MatrixMxN.h
```diff
@@ -1189,8 +1189,26 @@ class MatrixMxN
...
+        for (size_t i = 0; i < _MD; ++i)
```
asylvest: Avoid single character variable names (motivation is they're harder to do find/replaces on, though granted any decent IDE has a 'whole word' option).  So, these would change to `ii` and `jj`.

modules/c++/math.linear/unittests/test_mx.cpp
```diff
@@ -81,6 +81,50 @@ TEST_CASE(testScaleMultiplyMxN)
     TEST_ASSERT_EQ(B, C);
 }
 
+TEST_CASE(testNegateMxN)
+{
+    Matrix3x3 A = identityMatrix<3, double>();
+    Matrix3x3 B = -A;
+    Matrix3x3 C = A;
+    for (size_t i = 0; i < A.rows(); ++i)
```
asylvest: Ditto on `i` and `j` renames

modules/c++/math.linear/unittests/test_mx.cpp
```diff
@@ -81,6 +81,50 @@ TEST_CASE(testScaleMultiplyMxN)
     TEST_ASSERT_EQ(B, C);
 }
 
+TEST_CASE(testNegateMxN)
+{
+    Matrix3x3 A = identityMatrix<3, double>();
```
asylvest: To be pedantic, `A` and `B` can be `const`.

modules/c++/math.linear/unittests/test_mx.cpp
```diff
@@ -81,6 +81,50 @@ TEST_CASE(testScaleMultiplyMxN)
...
+    VectorN<3,double> Y;
+    X[0] = Y[2] =  1.;
+    X[1] = Y[1] =  0.;
+    X[2] = Y[0] = -1.;
+    TEST_ASSERT_EQ(X, -Y);
+}
+
+TEST_CASE(testNegate)
+{
+    Matrix2D<double> A = identityMatrix<3, double>();
```
asylvest: Ditto on `const`

modules/c++/math.linear/unittests/test_mx.cpp
```diff
@@ -81,6 +81,50 @@ TEST_CASE(testScaleMultiplyMxN)
...
+    Matrix2D<double> B = -A;
+    Matrix2D<double> C = A;
+    for (size_t i = 0; i < A.rows(); ++i)
```
asylvest: Ditto on single character variable names

modules/c++/math.linear/unittests/test_mx.cpp
```diff
@@ -560,6 +604,8 @@ int main()
     TEST_CHECK(testPtrAssign);
     TEST_CHECK(testSTLVectorAssign);
     TEST_CHECK(testOrthoTranspose5x5);
+    TEST_CHECK(testNegateMxN);
+    TEST_CHECK(testNegate);
```
asylvest: Can you add a `return 0` at the end of `main()`?  Looks like we forgot to do this in the past.  Also can get rid of the extra newlines that are in this function for whatever reason if you want.

asylvest: Can you add similar tests to `test_Vector`?

asylvest:
Code looks good... all just style nits

mwysockiMDA:
Okay, should be good to go.

PR #155, opened 2016-08-05
getCurrentExecutable() works by trying a number of different things o…

JonathanMeans:
…n Unix machines, so it doesn't work to throw if readLink fails

### DIFF
--- a/modules/c++/sys/source/OSUnix.cpp
+++ b/modules/c++/sys/source/OSUnix.cpp
@@ -56,7 +56,7 @@ std::string readLink(const std::string& pathname)
 
     if (readlink(pathname.c_str(), buffer, PATH_MAX) == -1)
     {
-        return "";
+        throw except::Exception(Ctxt(strerror(errno)));
     }
 
     return std::string(buffer);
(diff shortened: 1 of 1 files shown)


modules/c++/sys/source/OSUnix.cpp
```diff
@@ -56,7 +56,7 @@ std::string readLink(const std::string& pathname)
 
     if (readlink(pathname.c_str(), buffer, PATH_MAX) == -1)
     {
-        throw except::Exception(Ctxt(strerror(errno)));
+        return "";
     }
```
asylvest: Then do you want to change line 389 to...

` if (!executableName.empty() && isFile(executableName))`
?

Also, in what situation is there actually a symlink in `/proc/self/exe` that `readlink()` fails on?

JonathanMeans: I've changed it so it checks if the path exists before calling readlink. The issue I was encountering was that on Solaris the symlink is just elsewhere. Though a StackOverflow user did say that his system required root to follow the symlink in `/proc/self/exe`.

asylvest:
With the minor nit that I think personally the flow is a little easier to follow if you had enclosed the rest of the loop in `if (isFile(pathname))` rather than using `continue`, this logic looks good.  I like too that this way `readLink()` still throws rather than returning an empty string.

PR #156, opened 2016-08-08
Throw exception for invalid fit (not enough points)

mwysockiMDA:
math::poly::fit() throws if not enough points to constrain fit to unique solution; added unit tests for this.  For 1D case throw if npoints < order+1.  For 2D case throw if npoints < (orderX+1) * (orderY+1).  Essentially you need at least as many points as coefficients in your polynomial, or there are infinite solutions and the returned polynomial could be any of them (probably not what you want).

### DIFF
--- a/modules/c++/math.poly/include/math/poly/Fit.h
+++ b/modules/c++/math.poly/include/math/poly/Fit.h
@@ -6,6 +6,7 @@
 #include <math/linear/Matrix2D.h>
 #include <math/linear/VectorN.h>
 #include <numeric>
+#include <sstream>
 
 namespace math
 {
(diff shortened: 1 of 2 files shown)


modules/c++/math.poly/include/math/poly/Fit.h
```diff
@@ -6,6 +6,7 @@
 #include <math/linear/Matrix2D.h>
 #include <math/linear/VectorN.h>
 #include <numeric>
+#include <sstream>
```
asylvest: Even though you are almost certainly going to get these includes via other headers, should include sys/Conf.h and except/Exception.h since you're using them here.

modules/c++/math.poly/include/math/poly/Fit.h
```diff
@@ -1,11 +1,17 @@
 #ifndef __MATH_POLY_FIT_H__
 #define __MATH_POLY_FIT_H__
 
+// coda-oss
```
asylvest: Don't actually put comments like this saying where the includes are from.  I do like when they're grouped together with a newline between them like is done here though.

PR #158, opened 2016-08-23
Add Bessel function of the first kind

JonathanMeans:


modules/c++/math/source/Bessel.cpp
```diff
@@ -0,0 +1,175 @@
...
+            return besselIOrderN(order, x);
+    }
+}
+
+/*!
+ * Modified Bessel function of the first kind, order 0
+ */
+double besselIOrderZero(double x)
+{
+    double ax = std::fabs(x);
```
asylvest: Just use `std::abs()`

modules/c++/math/source/Bessel.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+    return ans;
+}
+
+/*!
+ * Modified Bessel function of the first kind, order 1
+ */
+double besselIOrderOne(double x)
+{
+    double ax = std::fabs(x);
```
asylvest: Use `std::abs()`

modules/c++/math/source/Bessel.cpp
```diff
@@ -0,0 +1,175 @@
...
+    const int IEXP = std::numeric_limits<double>::max_exponent / 2;
+    int k;
+    double bm, dum;
+
+    if (x * x <= 8.0 * std::numeric_limits<double>::min())
+    {
+        return 0;
+    }
+
+    double tox = 2.0 / std::fabs(x);
```
asylvest: ditto

modules/c++/math/source/Bessel.cpp
```diff
@@ -0,0 +1,175 @@
...
+        bip = bi;
+        bi = bim;
+        dum = std::frexp(bi, &k);
+
+        //Renormalize to prevent overflow
+        if (k > IEXP)
+        {
+            ans = std::ldexp(ans, -IEXP);
+            bi = std::ldexp(bi, -IEXP);
+            bip = ldexp(bip, -IEXP);
```
asylvest: Should be `std::ldexp()` for consistency?

modules/c++/math/source/Bessel.cpp
JonathanMeans: Fixed.

asylvest:
Looks good (I am certainly not double-checking the math but am assuming it's right).

PR #159, opened 2016-08-24
Add sign function

JonathanMeans:


modules/c++/math/include/math/Sign.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace math
+{
+/*!
+ * Find sign of input, expressed as -1, 0, or 1
+ * \param val A signed number
+ * \return 1 if val is positive, -1 if negative, 0 otherwise
+ */
+template <typename T>
+int sign(const T val)
```
asylvest: Passing in by value so no need for `const`

asylvest:
FYI to @mwysockiMDA... when you make an update to anything that has a Python binding, make sure to build on Linux and use the `--enable-swig` flag.  This will use Swig to automatically update the Python bindings for you (we check in the generated files so that when you build on Windows, or build without Swig, you still get Python bindings).  The addition of the `neg` operator in the Python bindings got committed on this branch.  For simplicity, no big deal... just FYI for next time.

modules/c++/math/include/math/Sign.h
JonathanMeans: Removed.

PR #160, opened 2016-08-31
Replacing PCRE 5.0 support with PCRE 10.22 support

asylvest:


asylvest:
@mwysockiMDA I've added in most of the way replacing PCRE 5.0 functionality with PCRE 10.22.  There are a few more waf configuration type things I wanted to sort out and test out.  One thing is that I'll put in something like an `--enable-std-regex` configure flag to trigger turning the STL version on (so that you can build with C++11 support but still choose which regex lib to use).  So don't get too hung up on the waf stuff, and I know you're not an expert in the PCRE2 syntax, so don't worry a ton about that either, but I wanted to get your general thoughts on the updates I made.  I did tweak a few interface things that affected RegexSTL.  I also tried to put a few more comments in the base Regex.h to explain what exactly the output vectors would contain, but I think my wording is klunky so feel free to propose something else.  I added a few more unit tests / expanded the existing tests a bit... if you have other things that you think should be tested to make sure there aren't any PCRE2 corners I didn't hit, again, feel free to suggest something.

One other thing I noticed... I got a compiler warning when building RegexSTL about an unused `matchBeginning` bool.  Sure enough, I see it get set to true initially and then false in subsequent loops but not used.  I didn't want to just remove it because I wasn't sure if it was supposed to be passed into that helper function and wasn't or if it just wasn't needed for the implementation.  Let me know.

It may be a little bit easier to look at the diff ignoring whitespace since I changed some indentations:
https://github.com/mdaus/coda-oss/pull/160/files?w=1

asylvest:
(Note that this is a pull request to merge my pcre2 branch into your stdregex branch... we don't have to literally do the pull request for this but I thought this was the easiest way to look at the diff and put in any comments)

mwysockiMDA:
Ugh, GitHub's diff algorithm is really bad.  Even ignoring whitespace it doesn't match-up the lines very well.  I'd ask for my money back, but...

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -26,37 +26,148 @@
...
+        const size_t subStringLength = end - index;
+        return str.substr(index, subStringLength);
+    }
+
+private:
+    // Noncopyable
+    ScopedMatchData(const ScopedMatchData& );
+    ScopedMatchData& operator=(const ScopedMatchData& );
+
+private:
```
mwysockiMDA: Extra private?

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -26,37 +26,148 @@
...
+    const pcre2_code* const mCode;
```
mwysockiMDA: I'm kind of surprised this works as const.  I would have guessed there were state variables that change.

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -70,244 +181,145 @@ re::Regex& re::Regex::operator=(const re::Regex& rhs)
...
+    std::string result = search(matchString, startIndex, 0, begin, end);
 
-    int idx = 0;
-    while (result.size() != 0)
+    while (!result.empty())
     {
         v.push_back(result);
-        idx += (int)result.size() + mOvector[0];
-        result = search(matchString, idx, PCRE_NOTBOL);
+        startIndex = end;
```
mwysockiMDA: Is `end` the last matching letter, or one beyond?  If it's the last matching, shouldn't this be `startIndex = end + 1;`?  Won't make a difference unless there are overlapping matches.  How many matches are there if we search for "aba" in "ababababa"?

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -70,244 +181,145 @@ re::Regex& re::Regex::operator=(const re::Regex& rhs)
...
+    size_t end;
+    size_t startIndex = 0;
+    std::string result = search(str, startIndex, 0, begin, end);
+    while (!result.empty())
     {
-        v.push_back(str.substr(idx, mOvector[0]));
-        idx += mOvector[1];
-        result = search(str, idx, PCRE_NOTBOL);
+        v.push_back(str.substr(startIndex, end - begin + 1));
+        startIndex = end;
```
mwysockiMDA: Same as above.

mwysockiMDA:
Okay, I have a small fix for the matchBeginning flag (yes it should be there, and yes I should be using it (derp)).  Should I commit to the stdregex branch or pcre2 branch?

modules/c++/re/source/RegexPCRE.cpp
asylvest: So I guess this isn't in the style guide per se, but in general I like to use extra `private`s like this to separate out sections... usually between private methods and private member variables

modules/c++/re/source/RegexPCRE.cpp
asylvest: I think `pcre2_code` is just used to get some basic info (all I know is `pcre2_match()` takes it in as const :o) )... note that `mMatchData` is only a `const` pointer, not a pointer to `const`, so that one presumably does change

modules/c++/re/source/RegexPCRE.cpp
asylvest: OK, easy part of the question first.  We use the terminology `[begin, end)` and `[first, last]` (the `[begin, end)` bit matches the STL's terminology).  But, the overlapping matches is a good call.  Actually, things are worse than that... take a look at the unit tests I just added.  The old functionality was not right; you can only increment `startIndex` to `begin + 1` safely (in your example, `startIndex = end` produced the wrong result already due to the overlap).  And actually, in making this change, the `testDotAllFlag()` function started failing which was because it wasn't right; the first test in there should have also found "3 4".  Now it does.

While we don't have the exact same condition in `sub()`, the PCRE implementation was still jumping a little too far.  Then `split()` wasn't handling the indexing for certain split patterns quite right either.  In both of these cases, you can't have overlapping patterns the same way you can in `searchAll()` at least.

Now thoroughly discouraged as it seems like all sorts of edge conditions were wrong, I added several more unittests to try to catch everything.  I think the PCRE implementation is good now, but see if you agree with the behavior of the unittests, any more edge conditions come to mind, and then also take a look at the STL implementation as it must have been off in its handling of some of these things too given that the unittests all passed previously for it.

This also means that the PCRE 5.0 implementation didn't correctly handle some of this either, so the good news I guess is we are fixing edge conditions in the code.  I'm certainly glad you took a look at these updates.

modules/c++/re/source/RegexPCRE.cpp
asylvest: Same deal with what `end` means, but based on the note below, something was incorrect in the indexing here anyhow... should be fixed now

asylvest:
Yeah so originally I made this pcre2 branch so that I could gradually check in pieces of pcre2 support as I had snippets of time and didn't want to leave things in a broken state on the stdregex branch in case you were doing something in there.  The remaining pcre stuff I need to do is all easy to do incrementally and keep the branch in a working state.  So I'd say for simplicity, let's both work in the pcre2 branch for the rest of this work now.  Hopefully we're in the home stretch of getting this into master.

modules/c++/re/unittests/test_regex.cpp
```diff
@@ -120,20 +202,75 @@ TEST_CASE(testMultilineBehavior)
...
+    re::Regex rx3("xxxxxxxxxx");
+    vec.clear();
+    rx3.split("ONExxxxxxxxxxTWOxxxxxxxxxxTHREE", vec);
+    TEST_ASSERT_EQ(vec.size(), 3);
+    TEST_ASSERT_EQ(vec[0], "ONE");
+    TEST_ASSERT_EQ(vec[1], "TWO");
+    TEST_ASSERT_EQ(vec[2], "THREE");
+
+    vec.clear();
+    rx3.split("ONExxxxxxxxxxTWOxxxxxxxxxxTHREExxxxxxxxxx", vec);
```
mwysockiMDA: So I noticed in split(), if the match is at the beginning, we return an empty string in the vector, but if the match is at the end, we don't.  I feel like these should behave the same way (I know in Python it includes the empty string for both).

asylvest:
Going to merge this into `stdregex` so that we have a single consolidated pull request we can reference

PR #166, opened 2016-09-06
Update lingering issues that only worked in Python2

JonathanMeans:
With this update, the requested tests will all pass with Python 3. Also replaces tabs with spaces.


asylvest:
Awesome - thanks

@clydestanfield

PR #169, opened 2016-09-08
Fix imports so Python 3 gets the correct pickle module.

JonathanMeans:
Test_math_poly.py now throws a SystemError with Python 3. I am almost certain this has nothing to do with pickle and will be resolved as soon as I get a SWIG upgrade. 

Also, msvs.py got switched back to using tabs completely to fix errors about mixed indentation when the template inside gets compiled.


JonathanMeans:
The SystemError is Python 3.5 specific. Works fine in 3.4

JonathanMeans:
After regenerated with SWIG 3.0.10, it works in 3.5 as well now.

asylvest:
Looks good.  Thanks for doing the Swig research to see that we needed to go to a newer version.

msvs.py is actually not ours... it's from waf.  If it does have some issues, we'll just have to be mindful of that when we upgrade to the latest version someday (we are now 3 years behind on the latest waf release I think).  But, no harm at this point in modifying the file when needed (looked like aside from whitespace changes, you added an import of sys?).

JonathanMeans:
So the key change for msvs.py was lines 355-360, to make it work with both versions of Python.

PR #170, opened 2016-09-12
Fix checking for SWIG version

JonathanMeans:
Two things here. I had to change the call to self.msg because option isn't a keyword argument. (docs: https://waf.io/apidocs/Context.html#waflib.Context.Context.msg)

And then it didn't actually do anything if it couldn't find the minimum version.


asylvest:
`swig.py` is provided by waf too but same deal... we'll just have to be mindful of this and verify it still works when we upgrade in the future.  It makes sense to make these kinds of fixes.

PR #171, opened 2016-09-13
Fix PCRE 10.22 linking on Windows

asylvest:
`PCRE2_STATIC` is not getting set in one additional spot which is causing linking issues on Windows


asylvest:
We built the PCRE library itself ok because this constant was in config.h but it needs to also be set prior to including pcre2.h, so generating it in re_config.h.

asylvest:
@mwysockiMDA

PR #172, opened 2016-09-20
Add package and makewheel commands for distribution 

JonathanMeans:
Distro branch in SIX is also ready.

### DIFF
--- a/build/makewheel.py
+++ b/build/makewheel.py
@@ -1,0 +1,30 @@
+from waflib.Build import BuildContext
+
+import glob
+import os
+import shutil
+import subprocess
+
+
+class makewheel(BuildContext):
+    '''builds a wheel for easy installation'''
+    cmd='makewheel'
+    fun='build'
+
+    def execute(self):
+        self.restore()
+        if not self.all_envs:
+            self.load_envs()
+
+        targetsStr = '--targets=' + self.targets
+
+        shutil.copyfile(os.path.join(self.getBuildDir(), 'six', 'conf',
+            'setup.py'), 'setup.py')
+        self.to_log('Creating wheel\n')
+        subprocess.call(['pip', 'wheel', '.', '--wheel-dir', '.'])
+        os.remove('setup.py')
+        numpyWheels = glob.glob('numpy*whl')
+        for wheel in numpyWheels:
+            os.remove(wheel)
+
+
--- a/build/makewheel.py
+++ b/build/makewheel.py
@@ -23,11 +23,6 @@ class makewheel(BuildContext):
         shutil.copyfile(os.path.join(self.env['SETUP_PY_DIR'],
             'setup.py'), 'setup.py')
         self.to_log('Creating wheel\n')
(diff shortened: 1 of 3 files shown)


build/makewheel.py
```diff
@@ -0,0 +1,27 @@
...
+    fun='build'
+
+    def execute(self):
+        self.restore()
+        if not self.all_envs:
+            self.load_envs()
+
+        targetsStr = '--targets=' + self.targets
+
+        shutil.copyfile(os.path.join(self.getBuildDir(), 'six', 'conf',
```
asylvest: This has `six` hardcoded in the path... need to generalize it somehow

build/makewheel.py
```diff
@@ -0,0 +1,27 @@
...
+    '''builds a wheel for easy installation'''
+    cmd='makewheel'
+    fun='build'
+
+    def execute(self):
+        self.restore()
+        if not self.all_envs:
+            self.load_envs()
+
+        targetsStr = '--targets=' + self.targets
```
asylvest: This isn't used for anything is it?

build/package.py
```diff
@@ -0,0 +1,32 @@
...
+        if not self.all_envs:
+            self.load_envs()
+
+        targetsStr = '--targets=' + self.targets
+
+        self.to_log('Zipping installation\n')
+        wheels = glob.glob('*.whl')
+        installDir = os.path.dirname(self.env['install_bindir'])
+
+        for wheel in wheels:
```
asylvest: Rather than needing to do this, should `waf package` just install right into `install_bindir` during `makewheel` above?

build/package.py
```diff
@@ -0,0 +1,32 @@
...
+    '''Creates a zip file of installation dir, and any wheels'''
+    cmd='package'
+    fun='build'
+
+    def execute(self):
+        self.restore()
+        if not self.all_envs:
+            self.load_envs()
+
+        targetsStr = '--targets=' + self.targets
```
asylvest: This isn't used for anything?

JonathanMeans:
These are fixed

JonathanMeans:
Er, never mind. Things are failing from a clean install somehow.

JonathanMeans:
This should be good to go again.

build/makewheel.py
```diff
@@ -0,0 +1,33 @@
...
+            raise Errors.WafError('Could not make wheel. '
+                'Unable to find setup.py')
+        shutil.copyfile(os.path.join(self.env['SETUP_PY_DIR'],
+            'setup.py'), 'setup.py')
+        self.to_log('Creating wheel\n')
+        subprocess.call(['pip', 'wheel', '.', '--wheel-dir', '.', '--no-deps'])
+        os.remove('setup.py')
+
+        wheels = glob.glob('*.whl')
+        for wheel in wheels:
```
asylvest: Do you have to do the move like this?  Can't you just set `--wheel-dir` above to `install_bindir`?  Or am I misunderstanding what that flag does?

JonathanMeans:
This is updated.

asylvest:
Code looks sane - will test further by integrating into SIX branch

PR #175, opened 2016-09-23
Regex url fix

clydestanfield:
I think this is correct?

I had to use the actual call in the net module to reproduce the issue. Using (abc)(d?)(efg) with "abcdefg" and "abcefg" would work as is. I am not regex savvy enough to parse everything going on in that call (I tried to put it into some online checkers and several of them failed on it).

This does work with the stl version of regex but that one always returns the maximum number of matches with extra empty strings at the end which is why line 81 is a check for GREATER rather than EQ. It looked like the code was setup to handle a variable number of returns so I assume PCRE is correct here (at least for our purposes). Do we want to change the stl version so it looks for trailing empty strings and removes them?

### DIFF
--- a/modules/c++/re/source/RegexPCRE.cpp
+++ b/modules/c++/re/source/RegexPCRE.cpp
@@ -103,7 +103,7 @@ public:
         }
         else
         {
-            return returnCode;
+            return pcre2_get_ovector_count(mMatchData);
         }
     }
 
@@ -114,11 +114,10 @@ public:
         const size_t index = outVector[idx * 2];
         const size_t end = outVector[idx * 2 + 1];
 
-        // If index is equal to end then we need to return an empty string.
-        // It is not always the case when this happens that these values will
-        // be less than or equal to the size of the string, so we need to
-        // explicitly return "" so nothing goes out of bounds.
-        if (index == end)
+        // If both index and end are set to PCRE2_UNSET then we did not find
+        // a match the index and end values are invalid for the substr call.
+        // In this case we need to return an empty string.
+        if (index == PCRE2_UNSET && end == PCRE2_UNSET)
         {
             return "";
         }
--- a/modules/c++/re/source/RegexPCRE.cpp
+++ b/modules/c++/re/source/RegexPCRE.cpp
@@ -83,7 +83,7 @@ public:
         // This returns the number of matches
         // But for no matches, it returns PCRE2_ERROR_NOMATCH
         // Other return codes less than 0 indicate an error
-        const int returnCode =
+        const int numMatches =
                 pcre2_match(mCode,
                             reinterpret_cast<PCRE2_SPTR>(subject.c_str()),
                             subject.length(),
@@ -103,6 +103,9 @@ public:
         }
         else
         {
+            // The num matches variable above won't include trailing empty
+            // matches. By returning the actual size including empty matches
+            // we now match the STL and Python versions of regex.
             return pcre2_get_ovector_count(mMatchData);
         }
     }
--- a/modules/c++/re/source/RegexPCRE.cpp
+++ b/modules/c++/re/source/RegexPCRE.cpp
@@ -114,6 +114,15 @@ public:
         const size_t index = outVector[idx * 2];
         const size_t end = outVector[idx * 2 + 1];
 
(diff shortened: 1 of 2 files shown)


mwysockiMDA:
This is interesting... if there is no subpattern match for a group, the pcre2 ovector indices will be PCRE2_UNSET (which is why you can add the `index==end` test).

The difference from the STL version is actually because we get the size from the return value of pcre2_match, which "is one more than the  highest numbered pair that has been set" and so won't include the empty matches at the end.  But if we call `pcre2_get_ovector_count(mMatchData)` to get the size instead, it returns the size _including_ the empty matches at the end (which really are there in the ovector).  I would propose changing the PCRE2 behavior to match the STL version, but then I'm always in favor of matching the Python regex behavior since that's what I'm familiar with :-)

modules/c++/re/unittests/test_regex.cpp
```diff
@@ -59,6 +59,33 @@ TEST_CASE(testMatches)
...
+    TEST_ASSERT_EQ(matches[2], "localhost");
+    TEST_ASSERT_EQ(matches[3], "80");
+    TEST_ASSERT_EQ(matches[4], "/something/page.com");
+    TEST_ASSERT_EQ(matches[5], "param1=foo&param2=bar");
+    TEST_ASSERT_EQ(matches[6], "fragment");
+
+    url = "http://localhost/page.com";
+    matches.clear();
+    rx.match(url, matches);
+    TEST_ASSERT_GREATER_EQ(matches.size(), 5);
```
asylvest: Add a comment about why this is - that the STL is behaving differently

modules/c++/re/unittests/test_regex.cpp
```diff
@@ -466,10 +493,12 @@ TEST_CASE(testHttpResponse)
     TEST_ASSERT_EQ(p.getContentLength(), "96");
 }
 
+
```
asylvest: Extra newline

asylvest:
Thanks for looking into this @mwysockiMDA.   I take it Python regex would also return the empty matches?  If that's the case, even though it seems weird to me that that's desirable, I begrudgingly agree that we should behave consistently with that.  So all we'd need to do is change line 106 of RegexPCRE.cpp to return `pcre2_get_ovector_count(mMatchData)`?

modules/c++/re/unittests/test_regex.cpp
asylvest: If we do change the behavior based on @mwysockiMDA's note, then scratch this... it should change to a `TEST_EQ()`

mwysockiMDA:
Yeah.  We could also add a getOutputVectorSize() member function for good measure.

My reasoning for why you would want this behavior is that if I make a regex with 7 subgroups, I would expect 7 sub-matches to those.  If I didn't want empty matches, I could always put a '?' after the group like `()?` (which is what the example in the unit tests does).

modules/c++/re/source/RegexPCRE.cpp
```diff
@@ -114,6 +114,15 @@ class ScopedMatchData
         const size_t index = outVector[idx * 2];
         const size_t end = outVector[idx * 2 + 1];
 
+        // If index is equal to end then we need to return an empty string.
+        // It is not always the case when this happens that these values will
+        // be less than or equal to the size of the string, so we need to
+        // explicitly return "" so nothing goes out of bounds.
+        if (index == end)
```
clydestanfield: Fixed.

Note, there are other situations where index == end and they do not equal PCRE2_UNSET. From what I can see, they are all correctly handled with the substr command though and return an empty string.

modules/c++/re/unittests/test_regex.cpp
clydestanfield: Updated based on Matt's feedback. PCRE now returns trailing empty strings. Changed to TEST_EQ.

modules/c++/re/unittests/test_regex.cpp
clydestanfield: Fixed

clydestanfield:
All the above comments have been addressed. PCRE and STL now match in the number of return matches (they both now return trailing empty empty strings).

asylvest:
Looks good - thanks guys

PR #176, opened 2016-09-24
Need to make HAVE_BOOST check smarter

asylvest:
Need to handle if modules/drivers/boost doesn't exist (because we've trimmed down the drivers directory in another repo)

@mbojrab


asylvest:
Specifically, if modules/drivers/boost/wscript doesn't exist, `bld.env['HAVE_BOOST']` won't be present, so it won't be false like the old way was counting on.

PR #177, opened 2016-09-26
Updated build.py recurse directories

clydestanfield:
Updated the recurse directories so drivers would always be first. This ensures things like libssh and AWS will be configured before checking for them in the env dictionary.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -60,9 +60,16 @@ class CPPContext(Context.Context):
             dirs = string
 
         if not dirs:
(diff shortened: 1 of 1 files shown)


asylvest:
FYI to @mbojrab - I think this would have automatically fixed the issue you ran into awhile ago where you weren't finding the FFTW driver before the fft module

FYI to @jmrandol @jbruce2112 for any C++ work you guys do

FYI to @msciaramitaro in case this affects any of your stuff (don't think it will)

PR #181, opened 2016-10-24
Matrix templates

JonathanMeans:
Make matrices of all dimensions used in SIX visible to SWIG.


modules/python/math.linear/source/math_linear.i
```diff
@@ -224,3 +237,1053 @@
         return returnVals;
     }
 }
+
+%extend math::linear::MatrixMxN<1, 1, double>
```
asylvest: Or... is it possible to do a `%extend` of the templated base class, add in these methods there, and then explicitly instantiate the templates via `%template` with that new extended class?

asylvest:
OK cool - this looks good

JonathanMeans:
I’ve updated it to use a macro.

Issue #173, opened 2016-09-22
makewheel flexibility


asylvest:
Currently makewheel finds the appropriate `setup.py` value by looking at `SETUP_PY_DIR`.  This works fine when there's exactly one wheel we want to make in a repo but doesn't work if we have a larger repo that potentially has multiple wheels that it might want to make.

Need to generalize this somehow... it seems like it would be nice to say something like
`waf makewheel --target=wheelName`, have associated that wheel name somehow in a wscript, and then have associated the path to setup.py based on this name.

JonathanMeans:
Can this be closed?

asylvest:
Yep - thanks

PR #183, opened 2016-10-26
Allow to specify targets for makewheel

JonathanMeans:
For Issue #173 . 

Regarding the changes to build.py: I find myself dreading figuring out how it works whenever it needs to be modified, so I'm trying to clean it up a little bit whenever I'm working in the area.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -69,7 +69,10 @@ class CPPContext(Context.Context):
             dirs.remove('drivers')
             dirs.insert(0, 'drivers')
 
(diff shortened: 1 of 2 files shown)


asylvest:
Thanks for the `build.py` cleanups - I agree it makes sense to keep inching those along to make them look more legible and the stuff you did is a step in the right direction.

I like the new wheel map - looks good.

FYI to @apakanati and @tolsonMDA - next time you grab a new CODA-OSS, the line in the wscript where you add the wheel target will change slightly.

PR #186, opened 2016-11-18
Don't define __CODA_CPP11 unless 11 is explicitly enabled

JonathanMeans:


asylvest:
Oh I see what was happening here... so I'm actually ok with the old behavior now that I understand this.  With gcc, we're explicit about if we want to turn C++11 support on or off since there's a separate compiler flag for it.  With Visual Studio, you just automatically get it if it's supported (the sanity check in `sys` is to see if we're using VS 2013 where we half-have C++11 support).

The problem that user was running into was really rooted in that we compiled with VS 2015 but he was trying to use with VS 2013 which wouldn't have worked regardless.  I don't think we need to artificially tie our hands behind our backs with Visual Studio, so let's leave it as-is.

PR #187, opened 2016-11-22
Add a RAII object for a temporary file

JonathanMeans:
I've kept needing these for testing, so I thought it would be convenient for CODA-OSS to just have one.


asylvest:
We have something similar in internal CODA... I think called `SafePath` in the `util` module.  Can you take a look to see how those two differ / if there's anything else we should support functionality-wise?  I think the `util` version uses UUID to generate a random name which is sort of an annoying dependency... rather than implementing the temporary filename stuff yourself, can't you use `OS::getTempName()` to simplify this?  Then in the destructor, you can use `OS::removeFile()`.  Put a try/catch around it as you can't throw an exception in a destructor without something horrible happening.

It would be good to allow the user to pass in an optional directory (it could default to the current working directory).

asylvest:
Since by using `sys::OS` there won't be anything system-specific in the implementation, what do you think about moving this to the `io` module?

JonathanMeans:
I've updated this. Also, I changed `OS::getTempName()` so it would end with a file being created in all cases.

modules/c++/io/include/io/TempFile.h
```diff
@@ -0,0 +1,48 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+#ifndef __TempFile_H__
```
asylvest: Include guard should have module name in it

modules/c++/io/include/io/TempFile.h
```diff
@@ -0,0 +1,48 @@
...
+namespace io
+{
+/*!
+ * RAII object for a temporary file that may either persist, or get deleted
+ * upon destruction
+ */
+class TempFile
+{
+public:
+    TempFile(bool destroy=true, const std::string& path=".");
```
asylvest: It seems like this class should always delete the file... if the caller didn't want that, they could just call `OS::getTempName()`

asylvest: Add Doxygen comments for the input parameters

asylvest: Rename `path` to `dirname` to be consistent with other code

asylvest: Spell out somewhere that the filename itself will be a random string or whatever... that is, that this class will pick it

modules/c++/io/include/io/TempFile.h
```diff
@@ -0,0 +1,48 @@
...
+    ~TempFile();
+    std::string pathname() const;
```
asylvest: Add Doxygen.  Might as well inline this method in the header

modules/c++/io/include/io/TempFile.h
```diff
@@ -0,0 +1,48 @@
...
+private:
+    bool mDestroy;
+    sys::OS mOS;
+    std::string mName;
+};
```
asylvest: Rename `mName` to `mPathname` for consistency

modules/c++/io/unittests/test_tempfile.cpp
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <fstream>
+
+#include <sys/OS.h>
```
asylvest: Don't need most of these includes.  Use angle brackets rather than quotes for all includes

asylvest: Whoops I meant this comment for TempFile.cpp not here

modules/c++/io/source/TempFile.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "io/TempFile.h"
+
+io::TempFile::TempFile(bool destroy, const std::string& path) :
+    mDestroy(destroy),
+    mOS(sys::OS()),
+    mName(mOS.getTempName(path))
+{
+    if (mName.empty())
+    {
+        throw except::Exception(Ctxt("Unable to create temporary file"));
```
asylvest: What do you think about having getTempName () throwing on failure rather than returning an empty string?  Then you don't need this here and the caller can always count on having a valid filename if it doesn't throw.

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -189,6 +189,7 @@ std::string sys::OSUnix::getTempName(const std::string& path,
     else
     {
         name = tempname;
+        fclose(fopen(name));
```
asylvest: The motivation is to make this consistent with Windows?  Could you just construct an unnamed `sys::File` here to get the same effect?

If creating the file is a side effect, update the AbstractOS documentation for this function.

JonathanMeans: Consistent with Window and Linux, yes. This path is only taken if we don't have Win32 or GCC. The other two cases create a file.

JonathanMeans:
This has been updated.

modules/c++/io/source/TempFile.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <import/except.h>
```
asylvest: You don't need the first two includes here

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -189,9 +189,14 @@ std::string sys::OSUnix::getTempName(const std::string& path,
     else
     {
         name = tempname;
+        sys::File newFile(name, sys::File::READ_AND_WRITE, sys::File::CREATE);
         free(tempname);
```
asylvest: You'll get a warning about an unused variable this way.  If you just leave it unnamed...

```
sys::File (name, ...);
```

You'll still get the side effect you want (constructor will create the file) without the warning.

asylvest: Unrelated but this is bugging me... in the `#if` section, we're using a fixed size buffer and some old school `strcat`s.  Can you change this to just use a `std::string`?

JonathanMeans: I'm not sure I can get rid of the buffer completely. mkstemp needs a char * that it can modify, and string.c_str() is const. I think the best I can do is to build the it up with std::string, then copy into the buffer before I pass it.

JonathanMeans:
Fixed.

modules/c++/io/include/io/TempFile.h
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * \return The pathname of the created file
+     */
+    inline std::string pathname() const
+    {
+        return mPathname;
+    }
+private:
+    sys::OS mOS;
+    std::string mPathname;
```
asylvest: Can be const

asylvest: This class shouldn't be copyable or assignable... can you put in private unimplemented methods for those like the mem classes have?

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -189,9 +188,14 @@ std::string sys::OSUnix::getTempName(const std::string& path,
     else
     {
         name = tempname;
+        sys::File (name, sys::File::READ_AND_WRITE, sys::File::CREATE);
```
asylvest: One more pedantic thing.  Things could throw here if either 1. The assignment into `name` fails (`std::string` ran out of memory) or 2. The file creation in `sys::File` fails.  In these scenarios, we will not get a chance to free `tempname` and it will leak.  Can you make an RAII class in the unnamed namespace that takes ownership of a `char*` and frees it in its destructor to guard against this?

Also technically the `sys::File` only needs to be opened with write permissions - don't need to read it

JonathanMeans:
Fixed again.

asylvest:
Looks good!

PR #190, opened 2016-12-12
Adding method to determine how many threads will be used by ThreadPlanner

asylvest:
Also adding Doxygen comments.

Moving some stuff from the header that doesn't need to be inlined into the cpp


chvink:
Looks good

PR #191, opened 2016-12-14
defaulting SWIG_PYTHON_EXTRA_NATIVE_CONTAINERS

apakanati:
This change is a global swig script change in order to change default SWIG behavior.  

In short, SWIG has an awesome default setting whereby depending on the number of elements in a vector, there is a covariant return type.

That is to say:
- a std::vector<T> of size 0 returns a tuple '()'
- a std::vector<T> of size 1 returns a T
- a std::vector<T> of size >1 returns a vector<T>

This introduces potentially really annoying handling at the python level.  I don't think that this is a good thing for us, and the benefit of whatever saving that might be is probably overwhelmed by the extra code maintenance.

After some digging, I came across this flag.  I think this is probably unintended behavior globally, so I've chosen to set a flag in swig.py unless -noextranative is explicitly passed:

If you're interested in what's going on at the preprocessor level, see below for SWIG 3.10 code:

  template <class Seq, class T = typename Seq::value_type >
  struct traits_from_stdseq {
    typedef Seq sequence;
    typedef T value_type;
    typedef typename Seq::size_type size_type;
    typedef typename sequence::const_iterator const_iterator;

    static PyObject *from(const sequence& seq) {
%#ifdef SWIG_PYTHON_EXTRA_NATIVE_CONTAINERS
      swig_type_info *desc = swig::type_info<sequence>();
      if (desc && desc->clientdata) {
    return SWIG_NewPointerObj(new sequence(seq), desc, SWIG_POINTER_OWN);
      }
%#endif
      size_type size = seq.size();
      if (size <= (size_type)INT_MAX) {
    PyObject *obj = PyTuple_New((Py_ssize_t)size);
    Py_ssize_t i = 0;
    for (const_iterator it = seq.begin(); it != seq.end(); ++it, ++i) {
      PyTuple_SetItem(obj,i,swig::from<value_type>(*it));
    }
    return obj;
      } else {
    PyErr_SetString(PyExc_OverflowError,"sequence size not valid in python");
    return NULL;
      }
    }
  };

The -extranative flag simply sets the define in generated CXX files so we always take the first option (e.g. a Vector<T>* regardless of its size.

### DIFF
--- a/build/swig.py
+++ b/build/swig.py
@@ -154,6 +154,8 @@ def i_file(self, node):
 
     flags = self.to_list(getattr(self, 'swig_flags', []))
     tsk.env.append_value('SWIGFLAGS', flags)
+    if not '-noextranative' in flags:
+        tsk.env.append_value('SWIGFLAGS', ['-extranative'])
 
     tsk.swig_install_fun = getattr(self, 'swig_install_fun', None)
 
--- a/modules/c++/numpyutils/examples/source/generated/numpyutilstest_wrap.cxx
+++ b/modules/c++/numpyutils/examples/source/generated/numpyutilstest_wrap.cxx
@@ -14,6 +14,7 @@
 #endif
 
 #define SWIG_PYTHON_DIRECTOR_NO_VTABLE
+#define SWIG_PYTHON_EXTRA_NATIVE_CONTAINERS
 
 
 #ifdef __cplusplus
--- a/modules/python/except/source/generated/coda_except_wrap.cxx
+++ b/modules/python/except/source/generated/coda_except_wrap.cxx
@@ -14,6 +14,7 @@
 #endif
 
 #define SWIG_PYTHON_DIRECTOR_NO_VTABLE
+#define SWIG_PYTHON_EXTRA_NATIVE_CONTAINERS
 
 
 #ifdef __cplusplus
--- a/modules/python/io/source/generated/coda_io_wrap.cxx
+++ b/modules/python/io/source/generated/coda_io_wrap.cxx
@@ -14,6 +14,7 @@
 #endif
 
 #define SWIG_PYTHON_DIRECTOR_NO_VTABLE
+#define SWIG_PYTHON_EXTRA_NATIVE_CONTAINERS
 
 
 #ifdef __cplusplus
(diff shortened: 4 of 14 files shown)


asylvest:
@chvink @JonathanMeans can you weigh in?

JonathanMeans:
I'm impressed you were able to track this flag down at all!

I noticed something while playing around with the update, and I'm not sure how concerning it is:

    >>> from coda.math_linear import *
    >>> innerVector = std_vector_double(3, 5.0)
    >>> outerVector = std_vector_vector_double(3, innerVector)
    >>> outerVector[0]
    <coda.math_linear.std_vector_double; proxy of <Swig Object of type 'std::vector< double > *' at 0x7f915ff54210> >
    >>> outerVector[0][1] = 1
    >>> outerVector[0][1]
    5.0
    >>> outerVector[0].push_back(3)
    >>> outerVector[0].size()
    3

Appending/modifying the outer vector works fine. Previously, the innerVector would have been a tuple, and then behaved as expected.

apakanati:
Yeah.... that seems crummy to me.  I'll need to spend some more time staring at this.

tolsonMDA:
I think Jon's observation is pretty concerning. As much as it's a bother that we might get non-vector returns, I think being unable to modify subvectors will probably cause more problems. As it is, there are two straightforward ways around the existing issue: either modify the vector in place instead of returning it, or add a check to the return, and if it's length is <= 1 then manually construct the vector. It's not pretty, but it could all be done in the interface file so the python user never sees the extra code.

JonathanMeans:
To clarify: we already can't modify the subvectors. It's a bit nicer now in that they're stored as tuples, not vectors, so when you try to modify it, Python raises an Exception instead of silently doing the wrong thing.

tolsonMDA:
Good point Jon. If you were to pass the `outerVector` into a swig-wrapped function, would swig allow the subvectors to be modified internally? Does that behavior change with Anu's revision?

asylvest:
As discussed with @apakanati , I don't like any of the options and expect to regret this someday, but let's move forward for now...  @JonathanMeans has verified the update does not appear to break downstream use cases such as the SIX Python bindings.

PR #192, opened 2017-01-06
Two minor gcc 6.2 / Alpine Linux fixes

asylvest:
`execl()` wanting a cast was a compiler warning
Not finding `linux/unistd.h` was an error


clydestanfield:
This looks good to me.

PR #194, opened 2017-01-18
Add nChooseK function

JonathanMeans:
This came up while I was working on the converter, and I thought it would be good here.

### DIFF
--- a/modules/c++/math/include/math/Utilities.h
+++ b/modules/c++/math/include/math/Utilities.h
@@ -20,8 +20,8 @@
  *
  */
 
-#ifndef __MATH_SIGN_H__
-#define __MATH_SIGN_H__
+#ifndef __MATH_UTILITIES_H__
+#define __MATH_UTILITIES_H__
 
 #include <stdlib.h>
 namespace math
@@ -53,11 +53,13 @@ inline double square(double val)
 
 /*
  * Calculate the binomial coefficient
+ * Be wary of the possibility of overflow from integer arithmetic.
+ *
  * \param n number of possibilities
  * \param k number of outcomes
  * \return n choose k
  */
-size_t nChooseK(size_t n, size_t k);
+sys::Uint64_T nChooseK(size_t n, size_t k);
 }
 
 #endif
--- a/modules/c++/math/source/Utilities.cpp
+++ b/modules/c++/math/source/Utilities.cpp
@@ -2,7 +2,7 @@
  * This file is part of math-c++
  * =========================================================================
  *
- * (C) Copyright 2004 - 2016, MDA Information Systems LLC
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
  *
  * math-c++ is free software; you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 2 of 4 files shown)


modules/c++/math/source/Utilities.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <import/sys.h>
```
asylvest: So I know we have the import headers, and I don't have anything stronger than "I don't like them and it feels good to include just the headers you need."  So in this case that would be `sys/Conf.h` and `except/Exception.h` I think.

modules/c++/math/source/Utilities.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <math/Utilities.h>
+
+namespace math
+{
+size_t nChooseK(size_t n, size_t k)
+{
+    if (n < k)
+    {
+        throw except::Exception(Ctxt("n Choose k undefined for n < k"));
```
asylvest: Add the values of `n` and `k` to the exception message.  That way if this somehow gets triggered 35 levels down of some complex chain and we're seeing this in a log message, we know what values were mistakenly sent in.

modules/c++/math/source/Utilities.cpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    // Algorithm to compute n Choose k without using factorials found here:
+    // http://csharphelper.com/blog/2014/08/
+    //         calculate-the-binomial-coefficient-n-choose-k-efficiently-in-c
+    size_t coefficient = 1;
```
asylvest: If we're on a 32-bit system (which nowadays is really rare but is still possible), a `size_t` is pretty small.  I think if you use a `sys::Uint64_T` throughout, you'll be safer.

There is some case where you'll overflow that too and would rather use a double.  Assuming you're using this for relatively small values for what you're doing right now, I think it's fine to leave it as an integer though to get a little more precision... if we ever hit an overflow case in the future, we can revisit.  Maybe for now just put a comment in the header cautioning the caller that we are doing integer arithmetic so it could overflow.

modules/c++/math/unittests/test_nChooseK.cpp
```diff
@@ -0,0 +1,55 @@
+/* =========================================================================
```
asylvest: The convention is snake_case for unit test filenames so this would be `test_n_choose_k.cpp`

modules/c++/math/unittests/test_nChooseK.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <TestCase.h>
+#include <math/Utilities.h>
+
+namespace
+{
+TEST_CASE(testNChooseK)
+{
+    TEST_ASSERT_EQ(math::nChooseK(1, 1), 1);
+    TEST_ASSERT_EQ(math::nChooseK(3, 2), 3);
+    TEST_ASSERT_EQ(math::nChooseK(10, 3), 120);
```
asylvest: What happens if I send 0 in?  Can you add a test for that?

modules/c++/math/include/math/Utilities.h
```diff
@@ -23,6 +23,7 @@
 #ifndef __MATH_SIGN_H__
```
asylvest: Ahh whoops - the include guard is goofed up

JonathanMeans:
This is ready.

modules/c++/math/source/Utilities.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <math/Utilities.h>
+
+namespace math
+{
+sys::Uint64_T nChooseK(size_t n, size_t k)
+{
+    if (n < k)
+    {
+        throw except::Exception(Ctxt("n Choose k undefined for n < k.\n"
+                "n: " + str::toString(n) + " k: " + str::toString(k)));
```
asylvest: Need to include a header for `str::toString()`

modules/c++/math/unittests/test_n_choose_k.cpp
```diff
@@ -0,0 +1,56 @@
...
+    TEST_ASSERT_EQ(math::nChooseK(10, 3), 120);
+}
+TEST_CASE(testNLessThanK)
+{
+    bool exceptionCaught = false;
+    try
+    {
+        math::nChooseK(3, 10);
+    }
+    catch(except::Exception& /*exception*/)
```
asylvest: OK total nit but you can catch this by const ref (and then, well... still do nothing with it)

JonathanMeans:
Fixed.

asylvest:
Looks good

PR #195, opened 2017-01-19
Add curl handle

clydestanfield:
This moves the functionality that was in the other repo into CODA-OSS to send curl commands.

There was no functionality change with this move, the only changes are:
* Everything is now part of net
* Added doxygen comments
* Added banners (updated to 2017).
* Added a driver for curl. Right now it only checks the system for a curl install as I don't think we have a use case to run this on Windows, but let me know if you want me to add a --with-curl-home option.
* Functionality is only enabled if NET_CURL_SUPPORT is set in net/net_config.h

@chvink


asylvest:
This looks good

PR #197, opened 2017-01-20
Prevent installation of python tests if not requested

JonathanMeans:


build/build.py
```diff
@@ -260,9 +260,13 @@ def module(self, **modArgs):
 
         pythonTestNode = path.parent.parent.make_node('python').make_node(str(path)).make_node('tests')
         if os.path.exists(pythonTestNode.abspath()) and not Options.options.libs_only:
-            for test in pythonTestNode.ant_glob('*.py'):
-                if str(test) not in listify(modArgs.get('test_filter', '')):
-                    self.install_files('${PREFIX}/tests/%s' % modArgs['name'], [test])
+            tests = [str(test) for test in pythonTestNode.ant_glob('*.py') if
+                    str(test) not in listify(modArgs.get('test_filter', ''))]
+            lib.targets_to_add.append(bld(features='install_tgt', pattern='**/*',
```
asylvest: So this does part of what we want.  But if I do `waf install --target=types-c++`, then I get the Python tests installed.  What I think should happen is either...

1. Only when you do `waf install --target=types-python` do these get installed
or
2. Only when you do `waf install --target=test_coda_types.py` does it get installed (i.e. it's not even added on as a target at all).

The second one would be more consistent with how our C++ tests work.  And in the case of SIX, just doing a `waf install` where it's going to walk the whole directory tree will still do what you want - all the Python tests under six/modules/python will still get installed.

JonathanMeans:
This is fixed to only install with the name of the Python test.

asylvest:
This looks like it's working great - thanks!

Issue #131, opened 2016-05-21
__init__.py doesn't always get installed


asylvest:
Still having intermittent issues with this... one theory I have is that in build.py's `python_package()`, `nod.sig` doesn't end up being unique when there are multiple Python modules because all of the `__init__.py` files are empty.  I had a point where I was trying to explicitly do `waf install --target=python_init_file_coda` and it wouldn't do anything.  Then I changed the `nod.sig` value to something unique and this worked.  But... the chaining still didn't: the module I was trying to install directly still didn't connect the `python_init_file_coda` target like it should have itself.

I wonder if we should just be creating this empty file once at configure time.  Then both the configure and build steps would be simpler... you wouldn't need the special `python_package` feature.

@chvink no need to work this right this second but keep it in the back of your head

@clydestanfield @apakanati @JonathanMeans

PR #198, opened 2017-01-20
Fix bug in creating __init__.py

JonathanMeans:
Fixes https://github.com/ngageoint/six-library/issues/43

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -1461,6 +1461,7 @@ def python_package(tg):
     if not os.path.isfile(fname):
         open(fname,'a').close()
 
+    if not os.path.isfile(os.path.join(install_path, os.path.basename(fname))):
         # to install files the 'node' associated with the file
         # needs to have a signature; the hash of the file is
         # good enough for us.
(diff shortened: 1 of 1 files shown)


asylvest:
I'm a little confused... how does this change fix the problem?  I thought `tg.bld.install_files()` only ran if the file didn't exist in the installation directory (due to some internal checking magic waf did), so can you explain how adding this check makes waf not get confused?  If I recall correctly, the issue we were seeing amounted to that sometimes waf got in a bad state and didn't re-generate the file... my best guess was due to getting the same signature (see notes on #131) though I could have been wrong.

JonathanMeans:
So the real problem was just above that, line 1461 in the latest commit. That line checks whether there's an __init__.py in the target directory, and all the installation stuff happened under that condition. So if you just removed the install dir, it would see the __init__.py still in target, and just not do anything. Removed the new check since it's redundant.

asylvest:
Ahh - ok!  I was just looking at the diff and just seeing whitespace changes and thinking what the heck.  Thanks for the explanation - that and looking at the full file view made clear to me that yeah, while technically "just whitespace changes", they had the effect of taking that stuff out of the conditional block.

OK, this makes sense.

asylvest:
FYI @chvink @clydestanfield

asylvest:
FYI @apakanati @tolsonMDA as you may have run into this before too

PR #199, opened 2017-01-23
Extract some repeated code into functions

JonathanMeans:


asylvest:
This is a good idea.  I haven't verified that everything still works properly after this change but I'm assuming you did. :o)

PR #200, opened 2017-01-24
Revert __init__.py change

JonathanMeans:
I am very occasionally, (maybe 1 time in 7?) getting an error while installing: ` '/data1/u/jmeans/coda/install/lib/python2.7/site-packages/coda/__init__.py' does not exist`

Haven't figured out how to reliably reproduce it yet. Until it's fixed, I think it's better to have __init__.py not be installed in certain, rare, known circumstances, than to have the install randomly fail.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -1438,15 +1438,15 @@ def python_package(tg):
     if not os.path.isfile(fname):
         open(fname,'a').close()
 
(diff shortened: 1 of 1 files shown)


asylvest:
Bummer. :o(

Yep I agree...

PR #201, opened 2017-01-24
 Only create one __init__.py per project, to prevent race conditions  

JonathanMeans:
I've verified through prints that python_package(), when called from SIX, only gets called twice (coda and pysix). If I'm right about what the problem is, this will fix it. Don't know how to be more sure, short of re-installing several dozen times.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -437,6 +437,7 @@ class CPPContext(Context.Context):
 
         return exe
 
+    _initHandled = []
     def swigModule(self, **modArgs):
         """
         Builds a SWIG C++ module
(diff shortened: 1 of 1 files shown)


JonathanMeans:
This is ready to go with `bld.get_tgen_by_name()`

build/build.py
```diff
@@ -499,10 +499,13 @@ def swigModule(self, **modArgs):
             # this turns the folder at the destination path into a package
 
             initTarget = init_tgen_name
-            bld(features = 'python_package',
-                name = initTarget,
-                target='__init__.py',
-                install_path = installPath)
+            try:
```
asylvest: OK cool - glad this worked.  This is going to be non-obvious when someone first looks at it (for example, in six months I will totally forget why we did it this way).  Can you add some comments explaining a) what get_tgen_by_name() is doing and why it's in a try/catch and b) what fundamentally we're doing here in terms of trying to get this target created exactly once but then reused across the targets with the same package name?

JonathanMeans: I've added the comments.

asylvest:
Looks good - thanks for getting this fixed!

FYI @apakanati

PR #202, opened 2017-01-25
Extract another function; resolve targets_to_add/targetsToAdd naming …

JonathanMeans:
…inconsistency


asylvest:
Cool - looks good.  I haven't verified things still build properly after this but I'm assuming you did and by inspection everything looks reasonable.  Thanks for these cleanups.

PR #203, opened 2017-01-26
Install tgt fix

clydestanfield:
I believe this will fix the Jenkins error we currently have.

It seems that install_tgt took in tsk as its first parameter instead of self. I updated the call to have it take in self instead. I am not sure what the original intent of the first parameter being tsk was, but I tested both "python waf install" and "python waf list" with "--install-source" and this seems to work.

@JonathanMeans


asylvest:
@JonathanMeans can you look at this?

JonathanMeans:
It looks correct, and I checked that CODA breaks before the change and works afterwards.

It looks like there's some kind of loose convention of having `tsk` as the first parameter for task generators, but it doesn't look like there'd be any reason to have it in this case.

PR #204, opened 2017-01-30
Some systems require explicitly adding curl to link line

chvink:
I ran into a case where one of our net libraries wouldn't compile because configuration couldn't implicitly link to libcurl, this will always explicitly link to it.

### DIFF
--- a/modules/drivers/curl/wscript
+++ b/modules/drivers/curl/wscript
@@ -8,6 +8,7 @@ def configure(conf):
     conf.env['HAVE_CURL'] = False
 
     if conf.check(uselib_store='CURL',
+                  lib='curl',
                   header_name='curl/curl.h',
                   function_name='curl_global_init',
                   msg='Checking for CURL',


asylvest:
Am I misremembering how waf works?  I thought you automatically got the name `curl` out of this?  What does waf do by default?

chvink:
I'm not sure how this works under the hood. I ran into a case earlier today where I couldn't build libraries that depended on the curl driver. The config.log error message told me that it couldn't find a symbol for `curl_global_init`, so I addded `lib='curl'` in the check to force waf to add the flag to the link line:

without `lib='curl'`:
`['/opt/gcc/4.9.1/bin/g++', '-fPIC', '-Wl,-E', '-m64', 'test.cpp.1.o', '-o', '/data1/u/cvink/progs/coda-oss/coda-oss/target/.conf_check_6be93fd46f300a63a7a991332a9516c5/testbuild/testprog', '-Wl,-Bstatic', '-Wl,-Bdynamic']`

with `lib='curl'`:
`['/opt/gcc/4.9.1/bin/g++', '-fPIC', '-Wl,-E', '-m64', 'test.cpp.1.o', '-o', '/data1/u/cvink/progs/coda-oss/coda-oss/target/.conf_check_4ebc058ea96e29d68eaeeea377dbce54/testbuild/testprog', '-Wl,-Bstatic', '-Wl,-Bdynamic', '-lcurl']`

As far as I can tell, `lib` is the only kwarg that actually tells waf that it needs to link to an additional library. The real puzzling part of this is why it works without `-lcurl` where it does.

asylvest:
Yeah hmm... I think that's what's confused me in the past is that sometimes it seems like we don't need it, but I agree that with other drivers, we explicitly put it in there.  So ok whatever - if it works, I'm happy.

PR #206, opened 2017-02-17
Undefine HAVE_CURL_GLOBAL_INIT

asylvest:
Otherwise it ends up in the waf environment variables


asylvest:
This is trivial and bugging me so merging it in...

PR #207, opened 2017-02-17
Typo in include guard

asylvest:
Doesn't look like this include guard was used elsewhere but it's still a mismatch.


clydestanfield:
This looks good. I think it can be merged.

PR #208, opened 2017-02-28
Add some safety checks when converting Polys to and from numpy arrays

JonathanMeans:


asylvest:
Good call - this all makes sense

PR #210, opened 2017-03-15
Move VectorString to coda_types

JonathanMeans:
I needed this template for other work, and thought things would work more nicely overall if it was with the others.


JonathanMeans:
This will have to get merged too for the SIX one to work.

PR #211, opened 2017-03-21
Updated handlers to avoid calling virtual methods in the destructor

clydestanfield:
This updates the StreamHandler to no longer call a virtual method from inside its destructor.

### DIFF
--- a/modules/c++/logging/source/StreamHandler.cpp
+++ b/modules/c++/logging/source/StreamHandler.cpp
@@ -40,7 +40,8 @@ StreamHandler::StreamHandler(LogLevel level) :
 
 StreamHandler::StreamHandler(io::OutputStream* stream,
                              LogLevel level) :
-    Handler(level)
+    Handler(level),
+    mClosed(false)
 {
     mStream.reset(stream);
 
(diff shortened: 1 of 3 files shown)


modules/c++/logging/source/StreamHandler.cpp
```diff
@@ -26,26 +26,40 @@
...
     // write prologue to stream
     write(mFormatter->getPrologue());
 }
 
-logging::StreamHandler::StreamHandler(io::OutputStream* stream,
-                                      logging::LogLevel level) :
-    logging::Handler(level)
+StreamHandler::StreamHandler(io::OutputStream* stream,
+                             LogLevel level) :
+    Handler(level)
```
asylvest: Need to add `mClosed` to the initializer list here

modules/c++/logging/source/StreamHandler.cpp
```diff
@@ -57,19 +71,30 @@ void logging::StreamHandler::setFormatter(logging::Formatter* formatter)
...
+
+void StreamHandler::closeImpl()
+{
+    if (!mClosed)
+    {
+        // end log with formatter injection
+        write(mFormatter->getEpilogue());
+
+        // delete formatter
+        Handler::close();
```
asylvest: But... this is still virtual

modules/c++/logging/source/StreamHandler.cpp
clydestanfield: Done

clydestanfield:
The comments have been addressed

PR #212, opened 2017-04-07
split() method can be const

asylvest:
You'll want to set aside some time to review this...


clydestanfield:
Looks good to merge.

PR #213, opened 2017-04-10
Always explicitly link to libcrypto if using net.ssl

chvink:
More systems than just solaris require this explicit linking, and it
doesn't seem to hurt systems where it's implicitly linked.


asylvest:
Does this work ok on Windows because it doesn't have `LIB_SSL`?

chvink:
Woops, now that I think about it I would guess not, but I forgot all about windows here. I'll update this PR as soon as I get to a machine to build it on.

chvink:
Just checked and this seems to work find on Windows.

asylvest:
OK cool - thanks for checking.

PR #216, opened 2017-04-17
Add isNaN function

JonathanMeans:


modules/c++/math/wscript
```diff
@@ -4,7 +4,11 @@ VERSION         = '0.1'
 USELIB          = 'MATH'
 MODULE_DEPS     = 'sys'
 
-options = configure = distclean = lambda p: None
+options = distclean = lambda p: None
+
+def configure(conf):
+    conf.check_cc(function_name='isnan', header_name='math.h',
```
asylvest: Do this like we do in the sys wscript where it makes a math_config.h file.  The reasoning is that that way someone building their application without waf  (j.e. just linking in our stuff) will have the defines they need.

JonathanMeans:
Good to go

modules/c++/math/include/math/Utilities.h
```diff
@@ -51,6 +52,24 @@ inline double square(double val)
...
+/*!
+ * Return true if argument is NaN
+ *
+ * \param value Argument to be checked for NaN
+ * \return true if value is NaN
+ */
+template <typename T> bool isNaN(T value)
+{
+#ifdef HAVE_ISNAN
+    return isnan(value);
```
asylvest: Include `math.h` to get the C `isnan()` function

modules/c++/math/wscript
```diff
@@ -1,10 +1,19 @@
...
 VERSION         = '0.1'
 USELIB          = 'MATH'
 MODULE_DEPS     = 'sys'
 
-options = configure = distclean = lambda p: None
+options = distclean = lambda p: None
+
+
+def configure(conf):
+    def sys_callback(conf):
```
asylvest: Rename to `math_callback()`

JonathanMeans:
Done.

modules/c++/math/include/math/Utilities.h
```diff
@@ -51,6 +52,25 @@ inline double square(double val)
...
+/*!
+ * Return true if argument is NaN
+ *
+ * \param value Argument to be checked for NaN
+ * \return true if value is NaN
+ */
+template <typename T> bool isNaN(T value)
+{
+#ifdef HAVE_ISNAN
+    #include <math.h>
```
asylvest: Put this up with the other includes.  I don't remember what happens if you include a header inside a namespace other than to say that there are times where it really really really doesn't do what you want, so you always want to put them up top.  For simplicity I wouldn't bother to put it inside a new ifdef up top; always including it is fine.

JonathanMeans:
It's moved.

asylvest:
Looks good!

PR #222, opened 2017-06-09
isnan -> std::isnan

JonathanMeans:


asylvest:
The `wscript` checks for `isnan()` from `math.h` rather than `std::isnan()` from `cmath`.  So, if we really need this, the `wscript` should be updated too.  But what is the need here... wasn't this working ok as-is?

JonathanMeans:
This is finally working.

modules/c++/include/TestCase.h
```diff
@@ -25,7 +25,6 @@
 
 #ifdef __cplusplus
 
-#  include <cmath>
```
asylvest: This was present in order to get `std::abs()` which is used in here... since you have the two sets of ifdefs, you can safely put this back in right?

modules/c++/math/include/math/Utilities.h
```diff
@@ -61,17 +61,20 @@ inline double square(double val)
  */
 template <typename T> bool isNaN(T value)
 {
-#ifdef HAVE_ISNAN
-    return isnan(value);
+#ifdef HAVE_STD_ISNAN
+    return std::isnan(value);
 #else
```
asylvest: You can simplify with `#elif`

modules/c++/math/source/Utilities.cpp
```diff
@@ -23,6 +23,7 @@
 #include <except/Exception.h>
 #include <str/Convert.h>
 #include <math/Utilities.h>
+#include <math.h>
```
asylvest: Not needed right?

modules/c++/math/wscript
```diff
@@ -11,7 +11,13 @@ options = distclean = lambda p: None
 def configure(conf):
     def math_callback(conf):
         conf.check_cc(function_name='isnan', header_name='math.h',
-                mandatory=False)
+                       mandatory=False)
```
asylvest: Extra space on line 14

modules/c++/math/wscript
```diff
@@ -11,7 +11,13 @@ options = distclean = lambda p: None
...
+        # The auto-generated code would be ambiguous and not compile
+        conf.check_cxx(fragment='''
+                       #include <cmath>
+                       int main() { return std::isnan(0); } ''', 
+                       msg='Checking for function std::isnan',
+		       mandatory=False, define_name='HAVE_STD_ISNAN')
```
asylvest: Need one more space on line 20

JonathanMeans:
Fixed.

modules/c++/math/include/math/Utilities.h
```diff
@@ -61,7 +61,9 @@ inline double square(double val)
  */
 template <typename T> bool isNaN(T value)
 {
-#ifdef HAVE_ISNAN
+#ifdef HAVE_STD_ISNAN
```
asylvest: Don't you need to include `cmath` if you have `HAVE_STD_ISNAN` vs. `math.h` if you have `HAVE_ISNAN`?

JonathanMeans: I've added the proper includes

asylvest: Including headers inside a namespace does bad things (it'll look like everything included in that header is inside the namespace) - https://stackoverflow.com/questions/2868955/shielding-include-within-namespace-block

So, you'll want to repeat the if/elif up at the top of the file rather than including them down here...

JonathanMeans: Fixed.

asylvest:
Looks good - that was easy. :o)

PR #223, opened 2017-06-19
Want a standalone utility that knows how to escape XML characters

asylvest:
We have a need for this functionality elsewhere... originally I was going to put it in xml.lite but then I saw that logging::XMLFormatter was actually already doing what I wanted to do (but could have been using `str::replaceAll()` to simplify it.  `logging` doesn't depend on `xml.lite`, so rather than add a dependency I put the functionality in `str` (which you could argue has no business knowing about XML but neither did `logging`).  I think it's more useful to have it here than it is to have two versions of the same code.  Added unittests for it.

### DIFF
--- a/modules/c++/str/source/Manip.cpp
+++ b/modules/c++/str/source/Manip.cpp
@@ -274,6 +274,7 @@ void escapeForXML(std::string& str)
     replaceAll(str, ">", "&gt;");
     replaceAll(str, "\"", "&quot;");
     replaceAll(str, "'", "&apos;");
-    replaceAll(str, "\n", " ");
+    replaceAll(str, "\n", "&#10;");
+    replaceAll(str, "\r", "&#13;");
 }
 }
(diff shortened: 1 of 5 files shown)


modules/c++/str/source/Manip.cpp
```diff
@@ -225,39 +255,25 @@ std::vector<std::string> str::split(const std::string& s,
...
 {
-    std::transform(s.begin(), s.end(), s.begin(), (int(*)(int)) tolowerCheck);
+    // & needs to be first or else it'll mess up the other characters that we
+    // replace
+    replaceAll(str, "&", "&amp;");
+    replaceAll(str, "<", "&lt;");
+    replaceAll(str, ">", "&gt;");
+    replaceAll(str, "\"", "&quot;");
+    replaceAll(str, "'", "&apos;");
+    replaceAll(str, "\n", " ");
```
asylvest: Good point - done

PR #224, opened 2017-06-27
Add function for computing pseudo-inverse

JonathanMeans:
`math::poly::fit()` and `math::linear::inverseLU()` aren't giving as good results as desired, so adding some tools to make it easy to do it myself.

### DIFF
--- a/modules/c++/math.linear/include/math/linear/Matrix2D.h
+++ b/modules/c++/math.linear/include/math/linear/Matrix2D.h
@@ -1334,27 +1334,27 @@ template<typename _T> inline
 }
 
 /*!
- * Calculate the left Moore-Penrose pseudo-inverse
+ * Calculate the left inverse
  * \param mx The matrix to find the inverse of. Mx is assumed to
  * have full row rank
- * \return Left pseudo-inverse
- * \throws if matrix is not pseudo-invertible
+ * \return Left inverse
+ * \throws if matrix is not left-invertible
  */
 template<typename _T> inline
-    Matrix2D<_T> leftPseudoInverse(const Matrix2D<_T>& mx)
+    Matrix2D<_T> leftInverse(const Matrix2D<_T>& mx)
 {
     return inverse(mx.transpose() * mx) * mx.transpose();
 }
 
 /*!
- * Calculate the right Moore-Penrose pseudo-inverse
+ * Calculate the right inverse
  * \param mx The matrix to find the inverse of. Mx is assumed to
  * have full column rank
- * \return Right pseudo-inverse
- * \throws if matrix is not pseudo-invertible
+ * \return Right inverse
+ * \throws if matrix is not right-invertible
  */
 template<typename _T> inline
-    Matrix2D<_T> rightPseudoInverse(const Matrix2D<_T>& mx)
+    Matrix2D<_T> rightInverse(const Matrix2D<_T>& mx)
 {
     return mx.transpose() * inverse(mx * mx.transpose());
 }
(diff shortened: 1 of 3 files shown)


asylvest:
@dstarinshak can you comment on the math here?

Also, how would I know if I should use the left or the right version?

JonathanMeans:
A matrix that has a left inverse will have at least as many rows as columns. 

I thought about just having a `pseudoInverse` function that does the right thing depending on the dimensions, but I'm worried about getting to the point where I'm trying to solve y = Ax, and A is wider than it is tall so it computes the right-inverse, and I silently get a garbage answer.

dstarinshak:
Good reference (scroll down to the "Matrices" section):
https://en.wikipedia.org/wiki/Inverse_element

Some takeways:

- An mxn matrix cannot have both a left and a right inverse
-- The left inverse *can* exist only if n > m
-- The right inverse *can* exist only if m > n
- An mxn matrix can be without both a left- and right-inverse if it's not "full rank"
- A general mxn matrix is guaranteed to have a Moore-Penrose pseudo-inverse (but this type of inverse isn't being implemented in this pull request)

To answer @asylvest: you would use the left-inverse if n > m and the right inverse if m > n.
An mxn matrix equaltion Ax = b means there are m equations and n unknowns.

When n > m, you have more equations than unknowns. There could be infinitely many solutions because you have more degrees of freedom (unknowns) than information to pin them down (equations). As long as A is full rank, you can use the left inverse to solve the equation for x. The solution x* = A_left_inv b is the "best fit solution" in that the vector Ax* gets you closer to b than any other vector x. The opposite goes for m > n: more equations than unknowns. Either you have exactly 1 solution, or none (i.e. too many equations trying to pin the degrees of freedom down that no solution can satisfy all of them at once). You would use the right inverse.

To answer @JonathanMeans: A user would get a garbage solution to Ax=b when A is a wide, short matrix (m > n) if the matrix B = A.transpose() * A is close to being singular. This is because you need to take the inverse of B to compute the left inverse. When B is close to singular (determinant close to zero, bad condition number, etc.) the inverse can blow up when computed numerically. This is tied to the "full rank" mentioned above. If A is full rank, then A.transpose() * A is guaranteed to be invertible, so you're okay computing the left inverse to solve Ax=b. Checking the rank of a matrix numerically can get cumbersome if the matrix is large and isn't something you want to do every time you attempt to perform and pseudo-inverse. (It usually involves singular value decomposition.) Probably better of just doing what you've implemented and allow Matrix2D throw an exception when you attempt to invert A.transpose() * A.

Long comment short: math looks fine to me.

modules/c++/math.linear/include/math/linear/Matrix2D.h
```diff
@@ -1329,9 +1329,35 @@ template<typename _T> inline
...
         // TODO Add 4x4
-    
+
     return inverseLU<_T>(mx);
 }
+
+/*!
+ * Calculate the left inverse
+ * \param mx The matrix to find the inverse of. Mx is assumed to
+ * have full row rank
```
asylvest: Can you spell out here and in the rightInverse() method what "full row rank" / "full column rank" means (i.e. more cols than rows, etc.)?

modules/c++/math.linear/unittests/test_inverse.cpp
```diff
@@ -0,0 +1,94 @@
...
+    xx(0, 0) = 1;
+    xx(0, 1) = 5;
+
+    // We want this to be able to solve linear equations
+    const Matrix2D<double> yy = xx * matrix;
+    TEST_ASSERT_EQ(yy * inverse, xx);
+}
+}
+
+int main(int argc, char** argv)
```
asylvest: To avoid warnings on Windows, change this to either
`(int , char** )` or `(int /*argc*/, char** /*argv*/)`.  There are lots of old programs that don't do this but let's start doing it for new ones to not add more warnings.

asylvest:
Looks good.  Thanks for clarifying on the math @dstarinshak.

PR #226, opened 2017-07-26
Add xml::lite::Speed and associated unittests

MikeBiancalanaMDA:
xml::lite::Speed makes creating XML documents a lot faster!

### DIFF
--- a/modules/c++/xml.lite/include/xml/lite/Speed.h
+++ b/modules/c++/xml.lite/include/xml/lite/Speed.h
@@ -2,7 +2,7 @@
  * This file is part of xml.lite-c++
  * =========================================================================
  *
- * (C) Copyright 2004 - 2014, MDA Information Systems LLC
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
  *
  * xml.lite-c++ is free software; you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
@@ -26,9 +26,9 @@
 #include <string>
 #include <vector>
 #include <map>
-#include <import/str.h>
-#include "xml/lite/Document.h"
-#include "xml/lite/Element.h"
+#include <str/Convert.h>
+#include <xml/lite/Document.h>
+#include <xml/lite/Element.h>
 
 namespace xml 
 {
@@ -76,19 +76,17 @@ namespace lite
  *  utility and shallow learning curve will excuse the introduction of a new
  *  syntax.
  */
-class Speed : public xml::lite::Document
+class Speed : public Document
 {
   public:
     /*!
-     * \func initRoot
      * \brief Creates the root element of the XML document.
      *
      * \param[in] rootName The name of the root element of the document.
      */
-    void initRoot(const std::string &rootName);
+    void initRoot(const std::string& rootName);
 
     /*!
-     * \func addRootAttribute
      * \brief Adds an XML attribute to the root element of the document.
      *
      * \param[in] attrname A string that is the name of the new attribute.
@@ -96,10 +94,9 @@ class Speed : public xml::lite::Document
      *                     If value is not a string it will be turned into one internally.
      */
     template <typename T>
-      void addRootAttribute(const std::string &attrname, T value);
+    void addRootAttribute(const std::string& attrname, const T& value);
 
     /*!
-     * \func setData
      * \brief Sets the value of an element.
      *
      * If the element does not exist, it is created.
@@ -114,24 +111,22 @@ class Speed : public xml::lite::Document
      *                  will be converted internally.
      */
     template <typename T>
-      void setData(const std::string &elname, const T &value);
+    void setData(const std::string& elname, const T& value);
 
     /*!
-     * \func addAttribute
      * \brief Adds an XML attribute to an element.
      *
      * \param[in] elname   The full name of the element being modified. See
      *                     setData for syntax.
      * \param[in] attrname The name of the attribute to add to element elname
-     * \param[in] value    The value of the attribute attrname.If it is not a
+     * \param[in] value    The value of the attribute attrname.  If it is not a
      *                     string, it will be converted internally.
      */
     template <typename T>
-      void addAttribute(const std::string &elname,
-                        const std::string &attrname,
-                        T value);
+    void addAttribute(const std::string& elname,
+                      const std::string& attrname,
+                      const T& value);
     /*!
-     * \func setNextData
      * \brief Creates and sets the next element with the name given at the path
      *        specified.
      *
@@ -153,10 +148,9 @@ class Speed : public xml::lite::Document
      * \return A string containing the full name of the new element.
      */
     template <typename T>
-      std::string setNextData(const std::string &elname, const T &value);
+    std::string setNextData(const std::string& elname, const T& value);
 
     /*!
-     * \func declareNamespace
      * \brief Declares an XML namespace on a given element.
      *
      * Note that this method merely declares a namespace. To set the namespace
@@ -167,12 +161,11 @@ class Speed : public xml::lite::Document
      * \param[in] nsPrefix The namespace prefix being declared.
      * \param[in] nsURI The URI associated with the namespace.
      */
-    void declareNamespace(const std::string &elname,
-                          const std::string &nsPrefix,
-                          const std::string &nsURI);
+    void declareNamespace(const std::string& elname,
+                          const std::string& nsPrefix,
+                          const std::string& nsURI);
 
     /*!
-     * \func setNamespace
      * \brief Sets a namespace prefix on an element and possibly all children.
      *
      * The recurse option is true by default.
@@ -184,23 +177,23 @@ class Speed : public xml::lite::Document
      * recursion enabled will perform a needless search of the existing nodes.
      *
      * \param[in] elname The full name of the element being modified.
-     * \param[in] nsPrefix 
+     * \param[in] nsPrefix The namespace prefix being set on the element(s).
+     * \param[in] recurse If true, sets the nsPrefix on all children of the
+     *                    element named by elname. Default is true.
      */
-    void setNamespace(const std::string &elname,
-                      const std::string &nsPrefix,
-                      const bool recurse=true);
+    void setNamespace(const std::string& elname,
+                      const std::string& nsPrefix,
+                      const bool recurse = true);
     
     /*!
-     * \func getData
      * \brief Gets the character data of an element
      *
      * \param elname The full name of the element.
      * \return A String containing the character data.
      */
-    std::string getData(const std::string &elname);
+    std::string getData(const std::string& elname);
 
     /*!
-     * \func fromDocument
      * \brief Set up this object to use an existing XML document
      *
      * This method is used to take an existing XML document and extend or
@@ -214,10 +207,9 @@ class Speed : public xml::lite::Document
      * \param steal Steal ownership of the root element of doc.
      *              Default is true.
      */
-    void fromDocument(xml::lite::Document *doc, bool steal=true);
+    void fromDocument(Document* doc, bool steal = true);
 
     /*!
-     * \func setCDATA
      * \brief Set the value of a CDATA element at at location
      *
      *  setCDATA takes a string value and inserts it at the given location
@@ -237,105 +229,105 @@ class Speed : public xml::lite::Document
      * \param elname The full name of the element.
      * \param cdata  The contents of the CDATA section.
      */
-    void setCDATA(const std::string &elname, const std::string &cdata);
+    void setCDATA(const std::string& elname, const std::string& cdata);
     // TODO: 
-    // setComment(const std::string &elname, const std::string &cdata);
-    // declareSchema(const std::string &elname,
-    //               const std::string &nsPrefix,
-    //               const std::string &nsURI);
+    // setComment(const std::string& elname, const std::string& cdata);
+    // declareSchema(const std::string& elname,
+    //               const std::string& nsPrefix,
+    //               const std::string& nsURI);
 
   protected:
-    /*! \func buildTOC
+    /*!
      *  \brief Build the table of contents map
      *
      */
-    void buildTOC(const std::string &parent="",
-                  xml::lite::Element const * element=NULL);
+    void buildTOC(const std::string& parent="",
+                  const Element* element=NULL);
 
     struct KeyParts
     {
-      std::string proper_name;
+      std::string properName;
       std::string parent;
       std::string child;
-      std::string child_name;
-      size_t      child_ord;
+      std::string childName;
+      size_t      childOrd;
     };
     
-    KeyParts vivify(const std::string &elname);
-    KeyParts extractParts(const std::string &longname);
+    KeyParts vivify(const std::string& elname);
+    KeyParts extractParts(const std::string& longname);
 
     //! The table of contents for the XML document.
-    std::map<std::string, xml::lite::Element*> mTOC;
+    std::map<std::string, Element*> mTOC;
     //! The namespaces known to exist in the document.
     std::map<std::string, std::string> mNamespaces;
 
 };
-} // end namespace "lite"
-} // end namespace "xml"
 //-----------------------------------------------------------------
-  template <typename T>
-void xml::lite::Speed::setData(const std::string &elname, const T &value)
+template <typename T>
+void Speed::setData(const std::string& elname, const T& value)
 {
-  std::string toc_key = elname;
+  std::string tocKey = elname;
   if (mTOC.count(elname) == 0)
   {
     // 0 and 1 are the only options for std::map.count()
-    xml::lite::Speed::KeyParts parts = vivify(elname);
-    toc_key = parts.proper_name;
+    KeyParts parts = vivify(elname);
+    tocKey = parts.properName;
   }
-  mTOC[toc_key]->setCharacterData(str::toString<T>(value));
+  mTOC[tocKey]->setCharacterData(str::toString<T>(value));
 
 };
 //-----------------------------------------------------------------
-  template <typename T>
-std::string xml::lite::Speed::setNextData(const std::string &elname, const T &value)
+template <typename T>
+std::string Speed::setNextData(const std::string& elname, const T& value)
 {
   // Take the hit of parsing the element name into parts.
-  xml::lite::Speed::KeyParts parts = extractParts(elname);
+  KeyParts parts = extractParts(elname);
   // Create the parents of this element if need be.
-  xml::lite::Speed::KeyParts parent_parts = vivify(parts.parent);
+  KeyParts parentarts = vivify(parts.parent);
 
   // Find out how many elements by this name there are right now.
-  std::vector<xml::lite::Element* > sibs;
-  mTOC[parts.parent]->getElementsByTagName(parts.child_name, sibs, false);
+  std::vector<Element* > sibs;
+  mTOC[parts.parent]->getElementsByTagName(parts.childName, sibs, false);
 
-  std::string next_name =
-    parts.parent+">"+parts.child_name+"&"+str::toString(sibs.size()+1);
-  xml::lite::Speed::KeyParts next_parts = vivify(next_name);
+  std::string nextName =
+    parts.parent+">"+parts.childName+"&"+str::toString(sibs.size()+1);
+  KeyParts nextParts = vivify(nextName);
 
-  mTOC[next_parts.proper_name]->setCharacterData(str::toString<T>(value));
+  mTOC[nextParts.properName]->setCharacterData(str::toString<T>(value));
 
-  return next_parts.proper_name;
+  return nextParts.properName;
 };
 //-----------------------------------------------------------------
-  template <typename T>
-void xml::lite::Speed::addAttribute(const std::string &elname,
-                             const std::string &attrname,
-                             T value)
+template <typename T>
+void Speed::addAttribute(const std::string& elname,
+                         const std::string& attrname,
+                         const T& value)
 {
-  std::string toc_key = elname;
+  std::string tocKey = elname;
   if (mTOC.count(elname) == 0)
   {
-    xml::lite::Speed::KeyParts parts = vivify(elname);
-    toc_key = parts.proper_name;
+    KeyParts parts = vivify(elname);
+    tocKey = parts.properName;
   }
-  xml::lite::Attributes attributes = mTOC[toc_key]->getAttributes();
-  xml::lite::AttributeNode node;
+  Attributes attributes = mTOC[tocKey]->getAttributes();
+  AttributeNode node;
   node.setLocalName(attrname);
   node.setValue(str::toString<T>(value));
   attributes.add(node);
-  mTOC[toc_key]->setAttributes(attributes);
+  mTOC[tocKey]->setAttributes(attributes);
 }
 //-----------------------------------------------------------------
-  template <typename T>
-void xml::lite::Speed::addRootAttribute(const std::string &attrname,
-                             T value)
+template <typename T>
+void Speed::addRootAttribute(const std::string& attrname,
+                                        const T& value)
 {
-  xml::lite::Attributes attributes = mRootNode->getAttributes();
-  xml::lite::AttributeNode node;
+  Attributes attributes = mRootNode->getAttributes();
+  AttributeNode node;
   node.setLocalName(attrname);
   node.setValue(str::toString<T>(value));
   attributes.add(node);
   mRootNode->setAttributes(attributes);
 }
+} // end namespace "lite"
+} // end namespace "xml"
 #endif
--- a/modules/c++/xml.lite/include/xml/lite/Speed.h
+++ b/modules/c++/xml.lite/include/xml/lite/Speed.h
@@ -1,0 +1,341 @@
+/* =========================================================================
+ * This file is part of xml.lite-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2014, MDA Information Systems LLC
+ *
+ * xml.lite-c++ is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef __XML_LITE_SPEED_H__
+#define __XML_LITE_SPEED_H__
+
+#include <string>
+#include <vector>
+#include <map>
+#include <import/str.h>
+#include "xml/lite/Document.h"
+#include "xml/lite/Element.h"
+
+namespace xml 
+{
+namespace lite
+{
+//=================================================================
+/*
+ * \class xml::lite::Speed
+ * \brief An efficient interface for creating XML documents
+ *
+ * This class is designed to make building XML documents simple and fast for
+ * developers. It features direct access to any element in the document and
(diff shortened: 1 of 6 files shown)


modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
+/* =========================================================================
+ * This file is part of xml.lite-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2014, MDA Information Systems LLC
```
asylvest: Update year to 2017

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+ *
+ */
+
+#ifndef __XML_LITE_SPEED_H__
+#define __XML_LITE_SPEED_H__
+
+#include <string>
+#include <vector>
+#include <map>
+#include <import/str.h>
```
asylvest: Just include `str/Convert.h` rather than import

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+#include "xml/lite/Document.h"
```
asylvest: Use angle brackets rather than quotes for these

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+ *  http://www.w3.org/TR/xquery-3). This class seeks to solve the problem of
+ *  creating an XML document where none exists, whereas the languages mentioned
+ *  above are focused on accessing and manipulating elements in pre-existing
+ *  XML documents. In order to facilitate autovivification the syntax must
+ *  unambiguously refer to single elements only, which would mean severely
+ *  restricting allowable XPath syntax for this use. It is hoped that the
+ *  utility and shallow learning curve will excuse the introduction of a new
+ *  syntax.
+ */
+class Speed : public xml::lite::Document
```
asylvest: So the name of this class should be a noun (well, I guess 'speed' is a noun but I think it needs to be more explicitly tied to the `Document`)... not sure what to suggest.  `SpeedDocument`?  `SpeedDoc`?  Is there something similar enough to XPath or XQuery that it's worth mentioning those?

asylvest: You're in the `xml.lite` namespace so can just inherit from `Document` rather than qualifying it

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+{
+  public:
+    /*!
+     * \func initRoot
```
asylvest: Thank you for putting Doxygen on all of these!

I think you can skip the `\func` everywhere though... won't Doxygen auto-figure this out?

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+     * \brief Creates the root element of the XML document.
+     *
+     * \param[in] rootName The name of the root element of the document.
+     */
+    void initRoot(const std::string &rootName);
```
asylvest: Much to my chagrin, CODA style is the ampersand/asterisk goes with the type, not with the variable (so `const std::string& rootName`)... applies throughout the code

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+    /*!
+     * \func addRootAttribute
+     * \brief Adds an XML attribute to the root element of the document.
+     *
+     * \param[in] attrname A string that is the name of the new attribute.
+     * \param[in] value    Value of the attribute.
+     *                     If value is not a string it will be turned into one internally.
+     */
+    template <typename T>
+      void addRootAttribute(const std::string &attrname, T value);
```
asylvest: Extra two spaces on this line (looks like that carries throughout the code.  the `v` in `void` should be right below the `t` in `template`)

asylvest: Might as well pass `value` in by const ref in case it's big right?

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+      void setData(const std::string &elname, const T &value);
+
+    /*!
+     * \func addAttribute
+     * \brief Adds an XML attribute to an element.
+     *
+     * \param[in] elname   The full name of the element being modified. See
+     *                     setData for syntax.
+     * \param[in] attrname The name of the attribute to add to element elname
+     * \param[in] value    The value of the attribute attrname.If it is not a
```
asylvest: Need a space or two after the period

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+     *                     string, it will be converted internally.
+     */
+    template <typename T>
+      void addAttribute(const std::string &elname,
+                        const std::string &attrname,
+                        T value);
```
asylvest: Pass in by const ref

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+
+    /*!
+     * \func addRootAttribute
+     * \brief Adds an XML attribute to the root element of the document.
+     *
+     * \param[in] attrname A string that is the name of the new attribute.
+     * \param[in] value    Value of the attribute.
+     *                     If value is not a string it will be turned into one internally.
+     */
+    template <typename T>
```
asylvest: Hmm, I thought there was a `\template` Doxygen tag that I was going to suggest putting on all these but... I guess there's not.

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+     * The element will be created if it does not already exist, but don't use
+     * this function for that. If there are going to be child nodes also in
+     * this namespace, consider creating them all using setData, then using
+     * setNamespace to set them all at once. Calling setNamespace for leaf
+     * nodes (as in the case where the element does not yet exist) with
+     * recursion enabled will perform a needless search of the existing nodes.
+     *
+     * \param[in] elname The full name of the element being modified.
+     * \param[in] nsPrefix 
+     */
```
asylvest: Add documentation for `recurse`.  Should have spaces between equals sign (`bool recurse = true`) and don't need `const` since you're passing it in by value

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+     * file and fill it out or possibly add to it. If the goal is gathering
+     * information for processing elsewhere, consider using XPath or XQuery,
+     * available in libxml2. CODA provides libxml2 as an optional driver for
+     * XML functionality.
+     *
+     * \param doc Pointer to the xml::lite::Document to now refer to.
+     * \param steal Steal ownership of the root element of doc.
+     *              Default is true.
+     */
+    void fromDocument(xml::lite::Document *doc, bool steal=true);
```
asylvest: Can just refer to this as `Document`

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+    //               const std::string &nsPrefix,
+    //               const std::string &nsURI);
+
+  protected:
+    /*! \func buildTOC
+     *  \brief Build the table of contents map
+     *
+     */
+    void buildTOC(const std::string &parent="",
+                  xml::lite::Element const * element=NULL);
```
asylvest: Can just refer to as `Element`.  `const` goes on the other side (`const Element* element`)

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+
+    struct KeyParts
+    {
+      std::string proper_name;
```
asylvest: camelCase for the variable names in here

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+     * \param cdata  The contents of the CDATA section.
+     */
+    void setCDATA(const std::string &elname, const std::string &cdata);
+    // TODO: 
+    // setComment(const std::string &elname, const std::string &cdata);
+    // declareSchema(const std::string &elname,
+    //               const std::string &nsPrefix,
+    //               const std::string &nsURI);
+
+  protected:
```
asylvest: If nothing realistically is ever going to inherit from this and get at this stuff, should be private rather than protected

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+    std::map<std::string, std::string> mNamespaces;
+
+};
+} // end namespace "lite"
+} // end namespace "xml"
+//-----------------------------------------------------------------
+  template <typename T>
+void xml::lite::Speed::setData(const std::string &elname, const T &value)
+{
+  std::string toc_key = elname;
```
asylvest: camelCase here too

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+  if (mTOC.count(elname) == 0)
+  {
+    // 0 and 1 are the only options for std::map.count()
+    xml::lite::Speed::KeyParts parts = vivify(elname);
```
asylvest: Can simply be `KeyParts` since you're in the class's method

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+    std::map<std::string, xml::lite::Element*> mTOC;
+    //! The namespaces known to exist in the document.
+    std::map<std::string, std::string> mNamespaces;
+
+};
+} // end namespace "lite"
+} // end namespace "xml"
+//-----------------------------------------------------------------
+  template <typename T>
+void xml::lite::Speed::setData(const std::string &elname, const T &value)
```
asylvest: Put all these methods inside the namespace rather than putting them outside... then you can remove all the `xml::lite::` business

modules/c++/xml.lite/include/xml/lite/Speed.h
```diff
@@ -0,0 +1,341 @@
...
+
+};
+//-----------------------------------------------------------------
+  template <typename T>
+std::string xml::lite::Speed::setNextData(const std::string &elname, const T &value)
+{
+  // Take the hit of parsing the element name into parts.
+  xml::lite::Speed::KeyParts parts = extractParts(elname);
+  // Create the parents of this element if need be.
+  xml::lite::Speed::KeyParts parent_parts = vivify(parts.parent);
```
asylvest: More snake_case... going to stop commenting on it :-P

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
+/* =========================================================================
+ * This file is part of xml.lite-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2014, MDA Information Systems LLC
```
asylvest: Update year

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <string>
+#include <vector>
+#include <map>
+#include <stdio.h>
+#include <stdlib.h>
+#include <import/str.h>
```
asylvest: Avoid these imports and just include what you need

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+#include <import/except.h>
+#include "xml/lite/Element.h"
```
asylvest: Angle brackets

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+//=================================================================
+void xml::lite::Speed::fromDocument(xml::lite::Document *doc, bool steal)
+{
+  setRootElement(doc->getRootElement(steal), steal);
+  buildTOC();
+}
+//=================================================================
+void xml::lite::Speed::buildTOC(const std::string &parent,
+                         xml::lite::Element const * element)
```
asylvest: Ditto comments in here regarding moving the const around

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+{
+  // make a smaller, local map for bookkeeping.
+  std::map<std::string, size_t> local_registry;
+  // either both or niether parent and element must be specified
```
asylvest: Typo - neither

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+  if (parent.empty() ^ (element == NULL))
+  {
+    throw except::Exception(
```
asylvest: Add `Ctxt()` inside all of these throws (and include `sys/Conf.h` up top).  This also buys you the method name which means this can basically change to just...

```
throw except::Exception(Ctxt("Called with only one of..."));
```

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+  if (element == NULL)
+  {
+    element = mRootNode;
+    // If we are starting at the top, clear out the old TOC before rebuilding.
+    mTOC.clear();
+  }
+
+  std::vector<xml::lite::Element *> childs = element->getChildren();
+  size_t sz = childs.size();
```
asylvest: You can assume this method is inlined and free, so just put it in the for loop condition

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+  for (size_t ii = 0; ii < sz; ii++)
+  {
+    std::string lname = childs[ii]->getLocalName();
```
asylvest: Can be `const`

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+{
+  xml::lite::Speed::KeyParts parts;
+  // make this name a proper name by ensuring zero-ordinal names have '&1' 
+  std::vector<std::string> chunks = str::split(longname, ">");
+
+  size_t ord_del_idx;
+  for(size_t ii = 0; ii<chunks.size(); ++ii)
+  { 
+    ord_del_idx = chunks[ii].find_last_of("&");
+    if (ord_del_idx == std::string::npos) chunks[ii] += "&1";
```
asylvest: Even one-liners always go on their own line with braces:

```
if (ord_del_idx == std::string::npos)
{
    chunks[ii] = "&1";
}
```

Also, I think you could skip having `ord_del_idx` at all and just put the `find_last_of` check on the left side of the condition... if you want to keep it, move the variable declaration inside the for loop since you don't need it elsewhere

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+  size_t delim_idx = parts.proper_name.find_last_of('>');
+  // grab string up to but not including the last '>' delimiter
+  if (delim_idx != std::string::npos)
+  {
+    parts.parent = parts.proper_name.substr(0, delim_idx);
+  }
+  else
+  {
+    parts.parent = "";
```
asylvest: Use `.clear()` instead

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+  parts.child_name = parts.child.substr(0, ord_delim_idx);
+  if (ord_delim_idx != std::string::npos)
+  {
+    parts.child_ord = atoi(parts.child.substr(ord_delim_idx+1).c_str());
+  }
+  else
+  {
+    // this branch should never happen- above code ensures presence of
+    // ordinal delimiter.
+    std::cout<<"NOT PROPER NAME! :"<<parts.child<<std::endl;
```
asylvest: If this should never happen, shouldn't this throw?  It shouldn't print... needs to either throw or we could pass in a logger if this is somehow just a warning

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+  xml::lite::Speed::KeyParts parts = extractParts(elname);
+  if (mTOC.count(parts.proper_name) == 0)
+  {
+    // No element of that name exists already
+
+    // TODO: Should we have an associated URI? Doesn't seem to make a
+    // difference.
+    xml::lite::Element* const child_elem = createElement(parts.child_name, "", "");
+
+    if (parts.parent.size() == 0)
```
asylvest: Use `.empty()` instead

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+//=================================================================
+void xml::lite::Speed::declareNamespace(const std::string &elname,
+                                 const std::string &nsPrefix,
+                                 const std::string &nsURI)
+{
+
+  std::string nsSep = ":";
+  // No prefix means default namespace
+  if (nsPrefix.size() == 0)
```
asylvest: Use `.empty()`

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+  {
+    nsSep = "";
+  }
+
+  if (elname.size() == 0)
```
asylvest: Use `.empty()`

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+  std::string toc_key = elname;
+  if (mTOC.count(elname) == 0)
+  {
+    // 0 and 1 are the only options for std::map.count()
+    xml::lite::Speed::KeyParts parts = vivify(elname);
+    toc_key = parts.proper_name;
+  }
+  mTOC[toc_key]->setCharacterData("<![CDATA["+cdata+"]]>");
+
+  return;
```
asylvest: Don't need this statement

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+    {
+
+      if ((toc_it->first).find(parts.proper_name) == 0
+          && toc_it->first != parts.proper_name)
+      {
+        setNamespace(toc_it->first, nsPrefix);
+      }
+    }
+  }
+  return;
```
asylvest: Don't need this statement

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
+/* =========================================================================
+ * This file is part of xml.lite-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2014, MDA Information Systems LLC
```
asylvest: Update year

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
+/* =========================================================================
+ * This file is part of xml.lite-c++
+ * =========================================================================
```
asylvest: File should be named `test_speed.cpp` (renamed when the class is renamed... lowercase snakecase for this file)

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+ *
+ */
+
+#include <string>
+#include <stdio.h>
+#include <import/io.h>
+#include <xml/lite/Speed.h>
+#include "TestCase.h"
+
+
```
asylvest: All of this stuff can be in an unnamed namespace

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+//#define __TEST_XML_LITE_SPEED_DEBUG__
+
+std::string writexml(const xml::lite::Speed &xml)
```
asylvest: Camelcase - either `writeXML()` or `writeXml()`

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <string>
+#include <stdio.h>
+#include <import/io.h>
```
asylvest: Avoid import

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <string>
+#include <stdio.h>
```
asylvest: Is this for `cout`?  Include `iostream`

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+  std::string out = ss_buf.stream().str();
+#ifdef  __TEST_XML_LITE_SPEED_DEBUG__
+  std::cout<<out<<std::endl;
+#endif
+
+  return out;
+}
+
+// Start with an empty root node called "root"
+#define SETUP_XML  xml::lite::Speed xml; xml.initRoot("root");
```
asylvest: Bah.  I really hate macros... especially this this is a two liner, I think it's worth putting explicitly in each test case.  Does it make sense to have, beyond the default constructor, a constructor that takes in the root?  Then this could be one line.

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+/////
+/// Begin Test Cases
+//
+//-----------------------------------------------------------------
+TEST_CASE(make_empty_root_node)
```
asylvest: camelcase for these and should have Test in the name - testMakeEmptyRootNode, etc.

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+  std::string truth = "<root><foo>yeah!</foo></root>";
+
+  TEST_ASSERT_EQ(writexml(xml), truth);
+}
+//-----------------------------------------------------------------
+TEST_CASE(child_data_type_conversion_ok)
+{
+  SETUP_XML;
+  xml.setData("foo", 3.14159f);;
```
asylvest: Extra semicolon

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+    if (parts.parent.size() == 0)
+    {
+      // requested element is at the top level (no parents)
+
+      // Ensure existence of lower ordinality siblings of the same name
+      if (parts.child_ord > 1)
+      {
+
+        std::string leftsib = 
+          parts.child_name+"&"+str::toString<int>(parts.child_ord - 1);
```
asylvest: Should be `size_t`, not `int`

modules/c++/xml.lite/source/Speed.cpp
```diff
@@ -0,0 +1,281 @@
...
+        // create parents which do not exist yet.
+        vivify(parts.parent);
+      }
+
+      // Ensure existence of lower ordinality siblings of the same name
+      if (parts.child_ord > 1)
+      {
+
+        std::string leftsib = 
+          parts.child_name+"&"+str::toString<int>(parts.child_ord - 1);
```
asylvest: Should be `size_t`, not `int`

modules/c++/xml.lite/unittests/test_Speed.cpp
```diff
@@ -0,0 +1,413 @@
...
+    "<A><one>1</one><three>3</three><two>2</two></A>"
+    "<lorem><ipsum><dolor><sit><amet>"
+    "consectetur adipiscing elit"
+    "</amet></sit></dolor></ipsum></lorem>"
+    "</root>";
+
+  TEST_ASSERT_EQ(writexml(foo), truth);
+}
+//=================================================================
+int main (int argc, char** argv)
```
asylvest: To avoid warnings about unused variables, either remove or comment out these names:
`int main(int, char**)`

modules/c++/xml.lite/include/xml/lite/Speed.h
MikeBiancalanaMDA: Yup, makes sense. Copyright is always a murky area for me.

modules/c++/xml.lite/include/xml/lite/Speed.h
MikeBiancalanaMDA: Ah yes, the old semantic war of "What are you really saying with where you put reference/pointer symbol?" Not a hill I'm willing to die on either. =]

modules/c++/xml.lite/include/xml/lite/Speed.h
MikeBiancalanaMDA: I agree with you and at the same time hope that attribute strings in XML are not huge. However, allowing this to play nicely in those inscrutable situations takes little effort, so I'll put this in.

As for the two spaces, I think vim was over-eager on the indentation. =]

modules/c++/xml.lite/include/xml/lite/Speed.h
MikeBiancalanaMDA: I actually do have a class that inherits from this- a specialization for KML. It is a discussion worth having whether inheritance or composition best suits that sort of usage though.

modules/c++/xml.lite/include/xml/lite/Speed.h
MikeBiancalanaMDA: > const goes on the other side (const Element* element)
Thank you! This looked wonky in going back through.

modules/c++/xml.lite/unittests/test_Speed.cpp
MikeBiancalanaMDA: The alternate constructor could make sense. I'm going to defer creating it since, that might encourage adding extra complexity, e.g. setting root attributes, namespaces, basic document structure, etc. in the constructor. Whether that functionality (and with it, complexity) is wanted can be decided at a later time, at the risk of that time never coming.

modules/c++/xml.lite/include/xml/lite/Speed.h
MikeBiancalanaMDA: You're welcome. =]

You're right about \func. Not only _can_ it be skipped, it _should_ be: https://www.doxygen.org/manual/commands.html#cmdfn

modules/c++/xml.lite/include/xml/lite/Speed.h
MikeBiancalanaMDA: Since I like the riff on "light speed" but do see that the class name is not clear as regards what the class actually _is_, I'll take your suggestion of SpeedDocument. I can't think of anything close to XPath or XQuery that I like.

MikeBiancalanaMDA:
@asylvest Thanks for taking the time to review this code. Your reading of it seems to have been thorough, which had to have taken a while.

PR #228, opened 2017-08-26
Guarding against dividing / modulo'ing by 0

asylvest:
If there are no elements per thread, you will simply use 0 threads.  This is way better than dividing / modulo'ing by 0 and getting everyone upset.


clydestanfield:
Looks good to me. I think this is safe to merge.

PR #229, opened 2017-09-14
Fix exception message for unrecognizable data type

JonathanMeans:
Previously, it would throw an exception here for trying to concatenate string and int.


asylvest:
Ahh whoops - thanks.

PR #231, opened 2017-10-26
add balanced runnables classes and interfaces to mt

christianwhite1193:


modules/c++/mt/include/mt/BalancedRunnable1D.h
```diff
@@ -0,0 +1,178 @@
...
+
+#include <sys/Conf.h>
+#include <sys/Runnable.h>
+#include <except/Exception.h>
+#include <mt/ThreadPlanner.h>
+#include <mt/ThreadGroup.h>
+
+namespace mt
+{
+/*!
```
asylvest: Needs a `\class` tag as well as `\tparam`

modules/c++/mt/include/mt/BalancedRunnable1D.h
```diff
@@ -0,0 +1,178 @@
...
+ *  and then increment an element, passing the fetched element to
+ *  the provided functor for processing. Each runnable will operate over the
+ *  full range of elements.
+ *
+ *  This runnable is useful in cases where work needs to be
+ *  done across a range of elements, but when dividing these elements
+ *  across threads leads to balancing issues i.e certain threads
+ *  terminate earlier than other threads.
+ *
+ *  \param[in,out] atomicCounter Atomic counter all threads will use to
```
asylvest: This section belongs down in the constructor

modules/c++/mt/include/mt/BalancedRunnable1D.h
```diff
@@ -0,0 +1,178 @@
...
+
+private:
+    const size_t mNumElements;
+    sys::AtomicCounter& mCounter;
+    const OpT& mOp;
+};
+
+/*!
+ *  For each thread, runs a BalancedRunnable1D object using the provided
+ *  functor.
```
asylvest: Need a `\tparam`.  Also, think of `BalancedRunnable1D` as implementation detail... in the sense that a normal user wants to look at just this function's documentation to get a full understanding of what this function does, rather than referencing the other runnable, so expand the description.

modules/c++/mt/include/mt/BalancedRunnable1D.h
```diff
@@ -0,0 +1,178 @@
...
+        {
+            threads.createThread(new BalancedRunnable1D<OpT>(
+                     numElements, counter, op));
+        }
+        threads.joinAll();
+    }
+}
+
+/*!
+ *  Same as above, but instead of sharing a functor across runnables,
```
asylvest: Add `\tparam`

modules/c++/mt/include/mt/BalancedRunnable1D.h
```diff
@@ -0,0 +1,178 @@
...
+            threads.createThread(new BalancedRunnable1D<OpT>(
+                    numElements, counter, ops[ii]));
+        }
+
+        threads.joinAll();
+    }
+}
+
+/*!
+ *  Convenience wrapper for providing each runnable with a copy of op.
```
asylvest: Add `\tparam`

modules/c++/mt/include/mt/WorkSharingBalancedRunnable1D.h
```diff
@@ -0,0 +1,295 @@
...
+#include <mt/ThreadPlanner.h>
+#include <mt/ThreadGroup.h>
+#include <types/Range.h>
+
+namespace mt
+{
+typedef std::vector<mem::SharedPtr<sys::AtomicCounter> > SharedAtomicCounterVec;
+
+/*!
+ *  This runnable takes both a reference to the associated thread's atomic
```
asylvest: Comments from previous file all apply to this file too

modules/c++/mt/include/mt/WorkSharingBalancedRunnable1D.h
```diff
@@ -0,0 +1,295 @@
...
+        mOp(op)
+    {
+    }
+
+    virtual void run()
+    {
+        // Operate over this thread's range
+        while (true)
+        {
+            const size_t element = mCounter.getThenIncrement();
```
asylvest: Not that this logic is super complicated, but we might as well not repeat it... the whole `while` block is repeated here and below.  Could have a private method that takes an atomic counter and an end element and does this while block, then call it in both spots

modules/c++/mt/include/mt/WorkSharingBalancedRunnable1D.h
```diff
@@ -0,0 +1,295 @@
...
+                                          ops[0]).run();
+    }
+    else
+    {
+        size_t threadNum = 0;
+        size_t startElement = 0;
+        size_t numElementsThisThread = 0;
+        const ThreadPlanner planner(numElements, numThreads);
+        std::vector<types::Range> threadPoolRange;
+        while(planner.getThreadInfo(
```
asylvest: Add space: `while (planner...`

modules/c++/mt/unittests/BalancedRunnable1DTest.cpp
```diff
@@ -0,0 +1,92 @@
+/* =========================================================================
```
asylvest: Application names are snake_case, so `balanced_runnable_1d_test.cpp`

modules/c++/mt/unittests/BalancedRunnable1DTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+TEST_CASE(BalancedRunnable1DTestWorkDoneSingleThreaded)
+{
+    const size_t numElements = 100000;
+    const size_t initValue = 0;
+    std::vector<size_t> workVec(numElements, initValue);
+
+    const size_t numThreads = 1;
```
asylvest: How about this step just steps through from one thread up to either 13 or `sys::OS().getNumCPUs()`?  Then you'd need just one test and you'd cover more cases in case there's a weird rounding bug.

modules/c++/mt/unittests/WorkSharingBalancedRunnable1DTest.cpp
```diff
@@ -0,0 +1,92 @@
+/* =========================================================================
+ * This file is part of mt-c++
```
asylvest: Ditto on filename change

modules/c++/mt/unittests/WorkSharingBalancedRunnable1DTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+TEST_CASE(WorkSharingBalancedRunnable1DTestWorkDoneSingleThreaded)
+{
+    const size_t numElements = 100000;
+    const size_t initValue = 0;
+    std::vector<size_t> workVec(numElements, initValue);
+
+    const size_t numThreads = 1;
+    IncOp op(workVec);
+
+    mt::runWorkSharingBalanced1D(numElements, numThreads, op);
```
asylvest: Ditto on for loop over num threads

modules/c++/types/include/types/Range.h
```diff
@@ -0,0 +1,91 @@
...
+#ifndef __TYPES_RANGE_H__
+#define __TYPES_RANGE_H__
+
+#include <limits>
+
+namespace types
+{
+/*!
+ * \class Range
+ * \brief Holds range information i.e a starting element and number elements
```
asylvest: missed "of" after "number"

christianwhite1193:
Updated according to comments.

modules/c++/mt/unittests/work_sharing_balanced_runnable_1d_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+        TEST_ASSERT_EQ(workVec.size(), numElements);
+
+        const size_t targetValue = 1;
+        for (size_t ii = 0; ii < numElements; ++ii)
+        {
+            const size_t value = workVec[ii];
+            TEST_ASSERT_EQ(value, targetValue);
+        }
+
+        std::fill(workVec.begin(), workVec.end(), initValue);
```
asylvest: Right now `workVec` only lives in the scope of the for loop so this line won't do anything useful.  Seems like you want to either move `workVec` out of the for loop and move this fill to the top or just remove the fill.

christianwhite1193: Fixed

PR #235, opened 2017-11-09
Add getNumSharedElements method to types::Range

christianwhite1193:


modules/c++/types/include/types/Range.h
```diff
@@ -86,6 +86,31 @@ struct Range
...
+                startElementToTest + numElementsToTest;
+
+        // Ranges do not intersect
+        if(mStartElement >= endElementToTest ||
+                endElement() <= startElementToTest)
+        {
+            return 0;
+        }
+
+        const size_t startRow = std::max(mStartElement, startElementToTest);
```
asylvest: include `<algorithm>` for max and min

modules/c++/types/include/types/Range.h
```diff
@@ -86,6 +86,31 @@ struct Range
                 (contains(startElement) && contains(startElement + numElements - 1)));
     }
 
+    /*!
+     * \param startElementToTest The start element
+     * \param numElementsToTest The total number of elements to check
+     *
+     * \returns The number of shared elements
+     */
+    size_t getNumSharedElements(size_t startElementToTest,
```
asylvest: Add a unit test for this (we don't have unittests for other methods of Range yet but that's ok... let's add a test_range.cpp with a unittest for just this one method as a start).

modules/c++/types/include/types/Range.h
christianwhite1193: header added

PR #236, opened 2017-11-11
runWorkSharingBalanced1DWithCopies() was calling itself

asylvest:
The compiler was getting stuck and eventually running out of memory... this is why. :o)

### DIFF
--- a/modules/c++/mt/include/mt/WorkSharingBalancedRunnable1D.h
+++ b/modules/c++/mt/include/mt/WorkSharingBalancedRunnable1D.h
@@ -307,7 +307,7 @@ void runWorkSharingBalanced1DWithCopies(size_t numElements,
                                         const OpT& op)
 {
     const std::vector<OpT> ops(numThreads, op);
-    runWorkSharingBalanced1DWithCopies(numElements, numThreads, ops);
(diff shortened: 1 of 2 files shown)


asylvest:
Found a second bug that occurred when the ThreadPlanner chose less threads than were initially provided (could be either because there were less items of work than threads or because of how ThreadPlanner rounds uneven amounts of work).  Fixed that too and added a unittest.

modules/c++/mt/include/mt/WorkSharingBalancedRunnable1D.h
```diff
@@ -271,8 +271,8 @@ void runWorkSharingBalanced1D(size_t numElements,
                       mem::SharedPtr<sys::AtomicCounter>(
                               new sys::AtomicCounter(startElement)));
 
-              threadPoolEndElements[threadNum++] =
```
christianwhite1193: It looks like `threadNum` needs to be incremented at the top of this loop in `planner.getThreadInfo(..`

asylvest: Oh jeez - thanks for catching this.  Fixed.

PR #237, opened 2017-11-20
Make io module easier to use

asylvest:
1. It is annoying to have to cast every buffer to a `sys::byte*` when reading and writing using `io::InputStream` and `io::OutputStream`.  Made this a `void*`.
2. Usually when you read, you really want all the bytes you asked for.  Added a new optional bool for if you want to throw on a short read.
3. Added some convenience functions to read the contents of a file in one shot.

Unfortunately 1 and 2 required touching a lot of code.

### DIFF
--- a/modules/c++/io/include/io/ReadUtils.h
+++ b/modules/c++/io/include/io/ReadUtils.h
@@ -30,11 +30,34 @@
 
 namespace io
 {
+/*!
+ * Reads the contents of a file (binary or text), putting the raw bytes in
+ * 'buffer'.  These are the exact bytes of the file, so text files will not
+ * contain a null terminator.
+ *
+ * \param pathname Pathname of the file to read in
+ * \param buffer Raw bytes of the file
+ */
 void readFileContents(const std::string& pathname,
                       std::vector<sys::byte>& buffer);
 
+/*!
+ * Reads the contents of a file into a string.  The file is assumed to be a
+ * text file.
+ *
+ * \param pathname Pathname of the file to read in
+ * \param[out] str Contents of the file
+ */
 void readFileContents(const std::string& pathname, std::string& str);
 
+/*!
+ * Reads the contents of a file into a string.  The file is assumed to be a
+ * text file.
+ *
+ * \param pathname Pathname of the file to read in
+ *
+ * \return Contents of the file
+ */
 inline
 std::string readFileContents(const std::string& pathname)
 {
--- a/modules/c++/io/include/io/ReadUtils.h
+++ b/modules/c++/io/include/io/ReadUtils.h
@@ -1,0 +1,47 @@
+/* =========================================================================
+ * This file is part of io-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
+ *
+ * io-c++ is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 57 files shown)


asylvest:
Oh, I cleaned up random warnings I saw when building along the way too

modules/c++/io/include/io/StringStream.h
```diff
@@ -128,6 +118,17 @@ class StringStream : public SeekableBidirectionalStream
         mData.clear();
     }
 
+protected:
+    /*!
+     * Read up to len bytes of data from this buffer into an array
+     * update the mark
+     * \param b   Buffer to read into
```
clydestanfield: Another missed "b" instead of "buffer"

modules/c++/io/source/ReadUtils.cpp
```diff
@@ -0,0 +1,52 @@
+/* =========================================================================
+ * This file is part of io-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
```
clydestanfield: Should this be Radiant Solutions?

modules/c++/io/source/ReadUtils.cpp
```diff
@@ -0,0 +1,52 @@
...
+    std::vector<sys::byte> buffer;
+    readFileContents(pathname, buffer);
+
+    if (buffer.empty())
+    {
+        str.clear();
+    }
+    else
+    {
+        str.assign(reinterpret_cast<char*>(&buffer[0]), buffer.size());
```
clydestanfield: This could be done without the extra copy by reading directly into str. If you are reading in an ASCII file you are probably not doing something super time critical so maybe it doesn't matter, but I just thought I should point it out.

modules/c++/io/tests/test_read_file_contents.cpp
```diff
@@ -0,0 +1,59 @@
+/* =========================================================================
+ * This file is part of io-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
```
clydestanfield: Another file that should be Radiant Solutions?

modules/c++/net/include/net/NetConnection.h
```diff
@@ -144,12 +135,21 @@ class NetConnection : public io::BidirectionalStream
...
-    virtual void write(const sys::byte* b, sys::Size_T len);
+    virtual void write(const void* buffer, size_t len);
 
     using io::BidirectionalStream::read;
     using io::BidirectionalStream::write;
 
 protected:
+    /*!
+     *  Read up to len bytes of data from input stream into an array
+     *  \param b   Buffer to read into
```
clydestanfield: Another "b" instead of "buffer"

modules/c++/io/include/io/ReadUtils.h
```diff
@@ -0,0 +1,47 @@
...
+#define __IO_READ_UTILS_H__
+
+#include <string>
+#include <vector>
+
+#include <sys/Conf.h>
+
+namespace io
+{
+void readFileContents(const std::string& pathname,
```
clydestanfield: It is probably worth adding doxygen to these functions. I imagine these are things people will want to use quite often.

modules/c++/io/include/io/ReadUtils.h
```diff
@@ -0,0 +1,47 @@
+/* =========================================================================
+ * This file is part of io-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
```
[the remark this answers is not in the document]
asylvest: Going to do this under a later PR

modules/c++/io/source/ReadUtils.cpp
asylvest: I believe you're not guaranteed that it's safe to write directly into the contents of a `std::string` (i.e. an implementation is free to use something other than a `char*` under the hood).  As you say, presumably this isn't a massive file... if someone wanted to avoid the copy, they could use the `vector<sys::byte` version.

PR #239, opened 2017-11-28
Avoiding namespace clash

asylvest:
Matching https://github.com/mdaus/nitro/pull/65
FYI @s-martin


asylvest:
@s-martin out of curiosity what was this clashing with?  I would have thought the compiler would have resolved it...

PR #240, opened 2017-12-13
Fix bugs when building drivers with --shared flag

JonathanMeans:
NRTAPI and NITFAPI are defined to be the appropriate __declspec statement.  Once this is merged, we will be able to do

```
python waf configure --prefix=install install
python waf configure --prefix=install --shared install


```and have DLLs for the C layer.

### DIFF
--- a/modules/drivers/zlib/wscript
+++ b/modules/drivers/zlib/wscript
@@ -110,7 +110,8 @@ def build(bld):
                       source=[os.path.join(minizipDir, 'ioapi.c'),
                               os.path.join(minizipDir, 'zip.c')],
                       path=driverNode,
-                      lib=os.path.join(os.path.join(os.getcwd(), env['install_libdir'], 'z')),
+                      lib='z',
+                      libpath=os.path.join(os.getcwd(), env['install_libdir']),
                       name='MINIZIP',
                       targets_to_add=[])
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/drivers/zlib/wscript [hunk 1]
```diff
@@ -103,36 +103,41 @@ def build(bld):
         # get compression to work, not decompression.
         minizipDir = os.path.join('contrib', 'minizip')
         minizipIncludes = ['.', minizipDir]
+        print('Building mijnizip')
```
asylvest: Looks like some debugging printouts left in here

modules/drivers/zlib/wscript
```diff
@@ -103,36 +103,41 @@ def build(bld):
...
         minizip = bld(features='c c%s add_targets' % env['LIB_TYPE'] or 'stlib',
                       includes=minizipIncludes,
                       export_includes=minizipIncludes,
                       target='minizip',
                       source=[os.path.join(minizipDir, 'ioapi.c'), 
                               os.path.join(minizipDir, 'zip.c')],
                       path=driverNode,
+                      lib=os.path.join(os.path.join(os.getcwd(), env['install_libdir'], 'z')),
```
asylvest: Why is this needed (or similarly, how did this work before)?

JonathanMeans: My understanding is that this tells you where to look for shared libraries that you're linking in. Since we weren't trying to build that before, it would not have caused problems.

asylvest:
Oh I see, minizip depends on z - gotcha.

PR #241, opened 2018-01-29
Adding header length constant

asylvest:


clydestanfield:
I think this is good to merge.

PR #246, opened 2018-03-08
Allowing access to single contiguous buffer in SwapBuffer

asylvest:
* Changing underlying allocation to be a single contiguous buffer
* Allowing direct access to this contiguous buffer for cases where you're at the very end of ping-ponging and need something that won't fit in just 'valid'... this way you don't need to allocate twice as much as you really need.

Debated about changing the second constructor to be just one buffer too but then there wasn't a way for the caller to guarantee alignment of 'scratch' if that was important and just in case there is a scenario where a caller does have two large separate buffers.  The downside is `getUnderlyingBuffer()` will be NULL in that case.  I'm open to other ideas.

FYI @dstarinshak

### DIFF
--- a/modules/c++/mem/include/mem/SwapBuffer.h
+++ b/modules/c++/mem/include/mem/SwapBuffer.h
@@ -29,7 +29,6 @@
 
 namespace mem
 {
-
 /*!
  *  \class SwapBuffer
  *  \brief Buffer for ping/pong processing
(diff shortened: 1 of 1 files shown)


asylvest:
Ended up not needing this - not merging it in order to avoid someone accidentally doing something goofy with a `SwapBuffer`

PR #248, opened 2018-03-13
Fixed Intel shared ptr

clydestanfield:
I was able to reproduce the error that @dstarinshak was seeing. This seems to fix it.

### DIFF
--- a/modules/c++/mem/include/mem/SharedPtrCpp11.h
+++ b/modules/c++/mem/include/mem/SharedPtrCpp11.h
@@ -43,7 +43,9 @@ template <class T>
 class SharedPtr : public std::shared_ptr<T>
 {
 public:
-    SharedPtr() = default;
+    SharedPtr() : std::shared_ptr<T>()
+    {
+    }
 
     using std::shared_ptr<T>::shared_ptr;
 


modules/c++/mem/include/mem/SharedPtrCpp11.h
```diff
@@ -43,7 +43,9 @@ template <class T>
 class SharedPtr : public std::shared_ptr<T>
 {
 public:
-    SharedPtr() = default;
+    SharedPtr() : std::shared_ptr<T>()
+    {
```
asylvest: So now you don't still need the `= default` line in order to automatically get other constructors right (because at this point we're just inheriting like any other class)?

For example, this still compiles on both Intel and gcc right:
```
int* foo = NULL;
mem::SharedPtr<int> fooPtr(foo);
```

clydestanfield: That is correct, using std::shared_ptr<T>::shared_ptr; gives you the other constructors. I know for sure that everything in test_shared_ptr.cpp compiles and runs successfully on both compilers.

asylvest: Cool beans

PR #249, opened 2018-04-06
Added equality/inequality operators to Range

asylvest:


clydestanfield:
This looks good to me

PR #252, opened 2018-04-24
Add ceilingDivide function; update Python

JonathanMeans:


modules/c++/math/include/math/Round.h
```diff
@@ -68,6 +70,23 @@ template<typename T> inline T round(T value, size_t fractionalDigits)
...
+
+/*!
+ * Equivalent to ceil((float)numerator / denominator)
+ *
+ * \param numerator Number to divide
+ * \param denominator Non-zero number to divide by
+ * \return Result of division, rounded up
+ * \throw if denominator is 0
+ */
+inline size_t ceilingDivide(size_t numerator, size_t denominator)
```
asylvest: Since I have written the equivalent of this function roughly four thousand times, thank you!

Since you're possibly throwing, this won't get inlined anyway, so put it in the cpp (and make sure to include sys/Conf.h and except/Exception.h)

Also, update mt::ThreadPlanner to use this in its mNumElementsPerThread calculation

modules/c++/math/unittests/test_round.cpp
```diff
@@ -157,12 +157,41 @@ TEST_CASE(testRoundDigits)
...
+    size_t d1 = 2;
+    TEST_ASSERT_EQ(math::ceilingDivide(n1, d1), 2);
+
+    size_t n2 = 5;
+    size_t d2 = 2;
+    TEST_ASSERT_EQ(math::ceilingDivide(n2, d2), 3);
+
+    size_t n3 = 1;
+    size_t d3 = 0;
+    try
```
asylvest: I think there's a macro that tests for throwing

modules/python/except/source/generated/coda_except.py
```diff
@@ -104,12 +104,9 @@ class Context(_object):
     __getattr__ = lambda self, name: _swig_getattr(self, Context, name)
     __repr__ = _swig_repr
 
-    def __init__(self, *args):
-        """
-        __init__(except::Context self, std::string const & file, int line, std::string const & func, std::string const & time, std::string const & message) -> Context
-        __init__(except::Context self, Context c) -> Context
-        """
-        this = _coda_except.new_Context(*args)
+    def __init__(self, file, line, func, time, message):
```
asylvest: Why'd all the new Swig stuff come in..?

JonathanMeans: Looks like it never got updated for https://github.com/mdaus/coda-oss/commit/6adcddf4013c4e383c5c0293507e7e4bc3773035

JonathanMeans: Would it be worthwhile to make Jenkins complain if it generates SWIG code different than what's checked in?

asylvest: If you can figure out an easy way to do that, that would be awesome.  I guess it's just checking if 'git diff' finds anything?

JonathanMeans:
Fixed

modules/c++/math/include/math/Round.h
```diff
@@ -24,6 +24,8 @@
 #define __MATH_ROUND_H__
 
 #include <cmath>
+#include <except/Exception.h>
```
asylvest: Don't need these two includes in header since the implementation is in the cpp now

PR #256, opened 2018-05-03
Exposing the UTCDateTime class to python

slipschutz:
Just exposes the sys/UTCDateTime.h object to python


asylvest:
Change itself looks good but you'll want to re-generate the Python bindings themselves via Swig (we check in the bindings).  So if you do `waf install --target=sys-python`, you should see two files in `modules/python/sys/source/generated` get updated; we want those too.  Make sure you're using Swig 3.0.12.

If you need a hand, @clydestanfield or @JonathanMeans can help.

slipschutz:
I have included the python bindings in the pull request

PR #259, opened 2018-05-17
Fix memory leak on error reading

JonathanMeans:
This memory leak triggers when `MinidomParser.parse()` throws, which I forced by giving it non-xml data. I'm uncertain if it's the same one they were seeing.

I agree that the destructor logic looks correct...

### DIFF
--- a/modules/c++/xml.lite/source/XMLReaderXerces.cpp
+++ b/modules/c++/xml.lite/source/XMLReaderXerces.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "xml/lite/XMLReader.h"
+#include <mem/ScopedArray.h>
 
 #if defined(USE_XERCES)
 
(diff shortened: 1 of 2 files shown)


modules/c++/xml.lite/source/XMLReaderXerces.cpp
```diff
@@ -39,18 +40,16 @@ void xml::lite::XMLReaderXerces::parse(io::InputStream & is, int size)
     {
         throw xml::lite::XMLParseException(Ctxt("No stream available"));
     }
-    sys::byte* buffer = new sys::byte[available];
-    oss.read(buffer, available);
+    mem::ScopedArray<sys::byte> buffer(new sys::byte[available]);
+    oss.read(buffer.get(), available);
 
     // Adopt my buffer, and delete it for me
```
asylvest: Old comment was wrong - update to `// Does not take ownership`

PR #260, opened 2018-05-22
Fix linker errors when building from tarball

JonathanMeans:
@dstarinshak  FYI

### DIFF
--- a/modules/drivers/j2k/openjpeg/wscript
+++ b/modules/drivers/j2k/openjpeg/wscript
@@ -19,127 +19,126 @@ def configure(conf):
 
     if j2kLayer == 'openjpeg':
         conf.env['MAKE_OPENJPEG'] = True
-        conf.env['HAVE_J2K']      = True
+        conf.env['HAVE_J2K'] = True
 
-    if j2kLayer == 'openjpeg' and not j2kHome:
+        if j2kHome:
+            conf.env['MAKE_J2K'] = False
+            conf.check(lib='openjp2 pthread', header_name='openjpeg.h',
+                       uselib_store='J2K',
+                       libpath=os.path.join(j2kHome, 'lib'),
+                       includes=os.path.join(j2kHome, 'include'),
+                       function_name='opj_start_compress',
+                       msg='Check for for openjpeg',
+                       okmsg=j2kHome,
+                       mandatory=True)
 
-        # callback function to check for #defines used by the openjpg module
-        def opj_callback(conf):
+        else:
+            # callback function to check for #defines used by the openjpg module
+            def opj_callback(conf):
+                # check functionality
+                conf.check_cc(header_name='stdint.h',
+                              define_name='OPJ_HAVE_STDINT_H',
+                              mandatory=False)
 
-            # check functionality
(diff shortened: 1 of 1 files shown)


modules/drivers/j2k/openjpeg/wscript
```diff
@@ -16,6 +16,11 @@ def options(opt):
 def configure(conf):
     j2kLayer = Options.options.j2k_layer
     j2kHome = Options.options.j2k_home
+
+    if j2kLayer == 'openjpeg':
```
asylvest: So I think this whole flow would be a little more clear (some of this pre-dates your changes) if it was...

```
if j2kLayer == 'openjpeg':
    HAVE_J2K = True
    MAKE_J2K = not j2kHome

    if MAKE_J2K:
        MAKE_OPENJPEG = True
        <existing stuff>
    else:
        <existing stuff>
```

modules/drivers/j2k/openjpeg/wscript
```diff
@@ -16,128 +16,129 @@ def options(opt):
...
-            conf.define('OPJ_VERSION_MINOR', '3', quote=False)
-            conf.define('OPJ_VERSION_BUILD', '0', quote=False)
-
-        # check for the source tarball
-        if not exists(join(conf.path.abspath(), SOURCE + '.tar')):
-            conf.fatal('Missing OpenJPEG tarfile')
-
-        # untar and setup env
+    if j2kLayer == 'openjpeg':
         conf.env['MAKE_OPENJPEG'] = True
```
asylvest: Right now `MAKE_OPENJPEG` is always getting set... but shouldn't it only be getting set if you didn't provide `j2kHome`?  Otherwise `build()` is going to try to build it right?

Also, you can simplify and just set `MAKE_J2K` as a function of having j2kHome so it's in one spot...

modules/drivers/j2k/openjpeg/wscript
JonathanMeans: I've simplified the logic. MAKE_J2K should not get set to False, because the parent wscript also uses it to decide whether to build Jasper. I just set it here because it makes all the environment checking look a little more explicit, and it won't ever be the case where `j2kLayer == 'openjpeg'`, `j2kHome` is set, and we still want to build Jasper.

modules/drivers/j2k/openjpeg/wscript
JonathanMeans: See modules/drivers/j2k/wscript:49

`MAKE_J2K` determines if we're building a J2K library at all. Then when we get into the specific wscript, `MAKE_OPENJPEG` and `MAKE_JASPER` are just verifying that we're building the one we're supposed to be building. Then `nitro/modules/c/j2k/wscript` uses `MAKE_OPENJPEG` and  `MAKE_JASPER` to check which library we have. 

If a prebuilt openjpeg is provided, we want MAKE_OPENJPEG set to say that we have openjpeg vs jasper, and MAKE_J2K=False (actually, just not set at all) to say that there's no need to actually build it.

JonathanMeans:
I'm not authorized to merge.

asylvest:
Whoops

PR #263, opened 2018-05-30
Adding Range::overlaps() convenience methods

asylvest:


modules/c++/types/include/types/Range.h
```diff
@@ -89,28 +89,52 @@ struct Range
...
-     * \returns The number of shared elements
+     * \return True if the ranges overlap, false otherwise
      */
-    size_t getNumSharedElements(size_t startElementToTest,
-                                size_t numElementsToTest) const
+    bool overlaps(const types::Range& rhs) const
     {
-        const size_t endElementToTest =
-                startElementToTest + numElementsToTest;
+        return (endElement() > rhs.mStartElement &&
```
clydestanfield: Should this by >= ? If your end element equals my start element, I would consider that to be overlapping a single element.

asylvest: End element is exclusive...  So if lhs `[start, end)` is `[100, 200)` and rhs is `[200, 300)`, they don't overlap

clydestanfield: Ok, then I agree this logic is correct. I think this is good to merge.

PR #267, opened 2018-09-14
Class to handle reservation of scratch memory segments with optional alignment

kdwilhelm:


modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef __MEM_SCRATCHMEMORY_H__
```
asylvest: camelCase counts as an underscore in the include guard, so should be `__MEM_SCRATCH_MEMORY_H__`

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+#include <except/Exception.h>
+#include <mem/BufferView.h>
+#include <sys/Conf.h>
+
+namespace mem
+{
+
```
asylvest: Extra newline

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+ *  Memory segments should be reserved during a "setup" phase using the put
+ *  method. Once all segments are reserved, call the setup method to set up
+ *  the underlying memory and ensure the alignment requirements of each segment.
+ *  The get method may be used afterwards to obtain pointers to the memory
+ *  segments.
+ */
+class ScratchMemory
+{
+public:
+    ScratchMemory() :
```
asylvest: Add `//! Default constructor`

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+        mIsSetup(false)
+    {
+    }
+
+    /*!
+     * \brief Reserve a buffer segment within this scratch memory buffer.
+     *
+     * \param key Identifier for scratch segment
+     * \param numBytes Size of scratch buffer
+     * \param numBuffers Number of distinct buffers to set up
```
asylvest: Put the defaults in the Doxygen comments.

modules/c++/mem/include/mem/ScratchMemory.h
clydestanfield: Add '_' between words when creating include guards. So change this to `__MEM_SCRATCH_MEMORY_H__`

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+     * \param key Identifier for scratch segment
+     * \param indexBuffer Index of distinct buffer
+     *
+     * \return Pointer to buffer segment
+     *
+     * \throws except::Exception if the scratch memory has not been set up,
+     *         the key does not exist, or index of buffer is out of bounds
+     */
+    template <typename T>
+    T* get(const std::string& key, size_t indexBuffer = 0);
```
asylvest: Is it ever useful to pass around a `const` `ScratchMemory` that can provide you `const T*`?  If it is, add a const overloading to `get()`.  If not, don't worry about it.

modules/c++/mem/include/mem/ScratchMemory.h
clydestanfield: Move the implementation for the constructor to the cpp file.

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+     * \param key Identifier for scratch segment
+     * \param numBytes Size of scratch buffer
+     * \param numBuffers Number of distinct buffers to set up
+     * \param alignment Number of bytes to align segment pointer
+     *
+     * \throws except::Exception if the given key has already been used
+     */
+    template <typename T>
+    void put(const std::string& key,
+             size_t numBytes,
```
clydestanfield: Rename this to numElements since you are going to multiply it by sizeof(T) later on.

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+    ScratchMemory(const ScratchMemory&);
+    ScratchMemory& operator=(const ScratchMemory&);
+
+    struct Segment
+    {
+        Segment(size_t numBytes, size_t numBuffers, size_t alignment) :
+            numBytes(numBytes),
+            numBuffers(numBuffers),
+            alignment(alignment)
+        {
```
clydestanfield: Move the implementation of the constructor to the cpp file.

modules/c++/mem/unittests/test_scratch_memory.cpp
```diff
@@ -0,0 +1,117 @@
...
+ *
+ */
+
+#include <mem/ScratchMemory.h>
+
+#include <mem/BufferView.h>
+#include <sys/Conf.h>
+#include "TestCase.h"
+
+namespace {
```
clydestanfield: Move the curly bracket down to the next line.

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -0,0 +1,140 @@
...
+        size_t numBuffers;
+        size_t alignment;
+        std::vector<sys::ubyte*> buffers;
+    };
+
+    std::map<std::string, Segment> mSegments;
+    std::vector<sys::ubyte> mStorage;
+    mem::BufferView<sys::ubyte> mBuffer;
+    size_t mNumBytesNeeded;
+    bool mIsSetup;
```
clydestanfield: Instead of having mIsSetup, can we just test if mBuffer.data == nullptr? That should test if setup has been called.

modules/c++/mem/include/mem/ScratchMemory.h
kdwilhelm: I think there could be situations where ScratchMemory is a member of some object and get is called indirectly using a const reference to that object. I added a const version using the Effective C++ method (implement the const version, delegate to it from the non-const version and cast away the const). Let me know if that is alright with you.

modules/c++/mem/include/mem/ScratchMemory.h
kdwilhelm: Done.

modules/c++/mem/include/mem/ScratchMemory.h
kdwilhelm: Done.

modules/c++/mem/include/mem/ScratchMemory.h
kdwilhelm: Done.

modules/c++/mem/include/mem/ScratchMemory.h
kdwilhelm: Done.

modules/c++/mem/unittests/test_scratch_memory.cpp
kdwilhelm: Done.

modules/c++/mem/include/mem/ScratchMemory.h
kdwilhelm: That's right. I also changed put to always set mBuffer.data to null, so get will throw in case the user calls put after setup without another call to setup.

modules/c++/mem/unittests/test_scratch_memory.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    // calling setup with buffer that is too small should throw
+    mem::BufferView<sys::ubyte> smallBuffer(buffer.data, buffer.size - 1);
+    TEST_EXCEPTION(scratch.setup(smallBuffer));
+
+    // calling setup with invalid external buffer should throw
+    mem::BufferView<sys::ubyte> invalidBuffer(NULL, buffer.size);
+    TEST_EXCEPTION(scratch.setup(invalidBuffer));
+}
+
```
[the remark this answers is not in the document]
kdwilhelm: Done.

modules/c++/mem/unittests/test_scratch_memory.cpp
```diff
@@ -0,0 +1,117 @@
...
+    // trying to get scratch before setting up should throw
+    TEST_EXCEPTION(scratch.get<sys::ubyte>("buf0"));
+
+    // set up external buffer
+    std::vector<sys::ubyte> storage(scratch.getNumBytes());
+    mem::BufferView<sys::ubyte> buffer(storage.data(), storage.size());
+
+    // first pass with external buffer, second with internal allocation
+    for (int ii = 0; ii < 2; ++ii)
```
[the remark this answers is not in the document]
kdwilhelm: Done.

modules/c++/mem/include/mem/ScratchMemory.hpp
```diff
@@ -0,0 +1,95 @@
...
+    mSegments.insert(
+            iterSeg,
+            std::make_pair(key, Segment(numElements, numBuffers, alignment)));
+}
+
+template <typename T>
+T* ScratchMemory::get(const std::string& key, size_t indexBuffer)
+{
+    return const_cast<T*>(
+            const_cast<const ScratchMemory&>(*this).get<T>(key, indexBuffer));
```
clydestanfield: Typically we try to avoid const cast where possible. What if instead of this we add a private method that handles the error checking and returns a `const Segment&`. We could then call that common method in both version and then only need a reinterpret_cast.

modules/c++/mem/include/mem/ScratchMemory.h
kdwilhelm: I reimplemented this as Clyde requested below:
https://github.com/mdaus/coda-oss/pull/267#discussion_r218024987

modules/c++/mem/include/mem/ScratchMemory.hpp
kdwilhelm: Done.

clydestanfield:
@asylvest This looks good to me. I think it is ready to merge unless you have any additional comments.

PR #268, opened 2018-09-18
Add convenience methods to get buffer view of scratch segment.

kdwilhelm:
A new method getBufferView helps to simplify usage of ScratchMemory in constructor initializer lists.

### DIFF
--- a/modules/c++/mem/include/mem/ScratchMemory.h
+++ b/modules/c++/mem/include/mem/ScratchMemory.h
@@ -108,8 +108,7 @@ public:
      *         the key does not exist, or index of buffer is out of bounds
      */
     template <typename T>
-    mem::BufferView<T> getBufferView(const std::string& key,
-                                     size_t indexBuffer = 0);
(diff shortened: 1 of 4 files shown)


asylvest:
Make sure to set Milestone field

modules/c++/mem/include/mem/ScratchMemory.h
```diff
@@ -96,6 +96,36 @@ class ScratchMemory
...
+     * \param key Identifier for scratch segment
+     * \param indexBuffer Index of distinct buffer. Defaults to 0.
+     *
+     * \return Buffer view of buffer segment
+     *
+     * \throws except::Exception if the scratch memory has not been set up,
+     *         the key does not exist, or index of buffer is out of bounds
+     */
+    template <typename T>
+    mem::BufferView<T> getBufferView(const std::string& key,
```
asylvest: Don't need to prefix these with `mem::` b/c you're in the `mem` namespace

PR #277, opened 2018-11-07
Fixed bug in nautical miles to feet conversion

asylvest:
We were multiplying by feet to meters rather than meters to feet


clydestanfield:
This looks good to me.

PR #279, opened 2018-11-19
ScratchVisualization

larislu:


modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
+#include <mem/ScratchMemory.h>
```
clydestanfield: Add copyright header

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+#include <cli/ArgumentParser.h>
+#include <str/Convert.h>
+#include <vector>
+#include <fstream>
+#include <string>
+
+struct Operation
+{
+    Operation(const std::string op,
+              const std::string name,
```
clydestanfield: Pass strings by const reference.

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+public:
+    /*!
+     * Constructor
+     *
+     * \param prevOperations All operations handled in previous iterations
+     * \param iteration Which iteration is being processed
+     * \param[in,out] htmlFile Ofstream to .html file
+     * \param[in,out] cssFile Ofstream to .css file
+     */
+    Visualizer(const std::vector<Operation> prevOperations,
```
clydestanfield: Pass vector by const reference

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+               const size_t iteration,
```
clydestanfield: Remove const from iteration. We don't use const when passing parameters by value.

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+        mReleasedColors.push_back("coral");
+        mReleasedColors.push_back("salmon");
+    }
+
+    /*!
+     * Find the lowest start address, which dictates the first block, and set mStartPtr.
+     *
+     * \param operations A vector of all operations
+     */
+    void set_start_ptr(const std::vector<Operation>& operations)
```
clydestanfield: Rename this to setStartPtr.

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+        }
+    }
+
+    /*!
+     * Write the CSS and HTML to create a box for a single operation.
+     *
+     * \param op Operation to draw a box of
+     * \param color_iter Iterator to arbitrarily select a color
+     */
+    void create_box(const Operation op, const size_t color_iter)
```
clydestanfield: Rename this to createBox. Pass op as const ref and remove const from color_iter. Rename color_iter to colorIter.

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+    void concurrentBlockTest(std::vector<Operation>& currentOperations,
+                             unsigned char& bufferName,
+                             size_t& testIter,
+                             size_t& usedBufferSpace,
+                             mem::ScratchMemory& scratch)
+    {
+        handlePrevOps(currentOperations, scratch);
+        size_t numElements = (rand() % 150) + 20;
+
+        if (testIter == 0)
```
clydestanfield: This looks like it is doing the same thing a `==1`. Can we change this to something like:
```
if (testIter == 0 || testIter == 1 | ...)
{
}
```

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+    void connectedBlockTest(std::vector<Operation>& currentOperations,
+                             unsigned char& bufferName,
+                             size_t& testIter,
+                             size_t& usedBufferSpace,
+                             mem::ScratchMemory& scratch)
+    {
+        handlePrevOps(currentOperations, scratch);
+        size_t numElements = (rand() % 150) + 20;
+
+        if (testIter == 0)
```
clydestanfield: Same thing here, combine like `if` statements.

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
+#include <mem/ScratchMemory.h>
+#include <mem/BufferView.h>
+#include <cli/ArgumentParser.h>
+#include <str/Convert.h>
+#include <vector>
+#include <fstream>
+#include <string>
+
+struct Operation
```
clydestanfield: Wrap everything here (upto main) in an unnamed namespace.

modules/c++/mem/tests/ScratchVisualization.cpp
```diff
@@ -0,0 +1,477 @@
...
+    std::ofstream& mCSSFile;
+};
+
+int main(int argc, char** argv)
+{
+    cli::ArgumentParser parser;
+
+    parser.setDescription("Software to visualize scratch memory test cases in HTML/CSS");
+    parser.addArgument("--test", "Select which test case to run", cli::STORE, "test")->setDefault("random");
+    parser.addArgument("--numBytes", "Determines how much memory is allocated for the buffer",
```
clydestanfield: Instead of --numBytes, can we just use the actual number of bytes from the ScratchMemory class? This would also show us that it allocates the size of buffer correctly.

modules/c++/mem/tests/ScratchVisualization.cpp
clydestanfield: Is this now duplicated here and in the other test code? It would be good to move this to a common location so that the unittests and this both run the same operations. Otherwise things may get out of sync.

larislu: What would be the best way to organize this? To put it as a header file in tests since the struct is only used for testing?

clydestanfield:
@asylvest I think this is good to merge.

PR #284, opened 2018-12-02
Adding Range operator<

asylvest:
This lets us use Range with an STL map


clydestanfield:
This looks good to merge.

PR #287, opened 2018-12-10
Adding SeekableNullOutputStream

asylvest:


asylvest:
Good call - re-generated Python bindings

PR #288, opened 2018-12-10
Update python bindings

JonathanMeans:
Apparently `waf clean` is required to get *everything*...


asylvest:
Interesting - thanks!

PR #291, opened 2019-02-08
Fix vector index issue

kdwilhelm:
Changed `&mBuffer[mBufferEnd]` to `mBuffer.data() + mBufferEnd` to avoid indexing the element past the end of a vector when only its address is needed.

### DIFF
--- a/modules/c++/io/include/io/StreamSplitter.h
+++ b/modules/c++/io/include/io/StreamSplitter.h
@@ -98,7 +98,7 @@ private:
      *        to the substring and remove it from the buffer.
      */
     void transferBufferSegmentToSubstring(std::string& substring,
-                                          size_t& tokenSize,
+                                          size_t& substringSize,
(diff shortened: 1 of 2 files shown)


asylvest:
`.data()` is C++11 which we're avoiding using in CODA-OSS (still :o( ).

You can do something like `mBuffer.empty() ? NULL : &mBuffer[0]` to get the initial pointer safely, though I didn't look closely enough at all the pointer arithmetic you're doing to see exactly how that would apply to your use case

kdwilhelm:
I'll try something like that. The issue was with calling std::copy, on a range including the last element in the buffer. In that case the end pointer passed to std::copy should be 1 past the end of the buffer, which cannot safely be obtained by indexing it and then taking its address.

The original implementation worked fine on Linux/gcc and passed Valgrind memcheck. I just caught in on Windows in debug mode, which enabled bounds checking for vector indexing.

PR #293, opened 2019-02-21
Mem Module Updates to Support Swig

christianwhite1193:


modules/c++/mem/include/mem/SharedPtr.h
```diff
@@ -25,6 +25,12 @@
 
 #include <sys/sys_config.h>
 
+namespace mem
+{
+template<typename T>
```
asylvest: Can you add a comment here explaining why you're doing this  Otherwise we'll forget in three months and delete this.

christianwhite1193: Comment added.

PR #296, opened 2019-03-12
Only add explicit bool if we have C++11

JonathanMeans:
GCC 4.4.7 gave a compile error here.

### DIFF
--- a/modules/c++/mem/include/mem/ScopedCopyablePtr.h
+++ b/modules/c++/mem/include/mem/ScopedCopyablePtr.h
@@ -25,6 +25,7 @@
 
 #include <memory>
 #include <cstddef>
+#include <sys/sys_config.h>
 
 namespace mem
 {
@@ -107,7 +108,7 @@ public:
         return !(*this == rhs);
     }
 
-    // explicit operators not suppored until C++11
+    // explicit operators not supported until C++11
     bool toBool() const
     {
         return get() == NULL ? false : true;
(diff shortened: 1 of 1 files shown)


modules/c++/mem/include/mem/ScopedCopyablePtr.h
```diff
@@ -106,11 +106,18 @@ class ScopedCopyablePtr
     {
         return !(*this == rhs);
     }
-    
-    explicit operator bool() const 
-    { 
-        return get() == NULL ? false : true; 
+
+    // explicit operators not suppored until C++11
```
asylvest: suppored --> supported

modules/c++/mem/include/mem/ScopedCopyablePtr.h
```diff
@@ -106,11 +106,18 @@ class ScopedCopyablePtr
...
+    bool toBool() const
+    {
+        return get() == NULL ? false : true;
     }
+#ifdef __CODA_CPP11
```
asylvest: You need to add one of the _config headers to get this

gpotts: For non CPP11 this should be valid

#ifdef __CODA_CPP11
    explicit operator bool() const
    {
        return toBool();
    }
#else
    operator bool() const
    {
        return toBool();
    }
#endif

asylvest: OK so in that case @JonathanMeans can you put out a new PR that collapses this to...
<remove `toBool()`>

```
#ifdef _CODA_CPP11
explicit
#endif
operator bool() const
{
    return get() == NULL ? false : true;
}
```

PR #299, opened 2019-03-18
Fixed failure to account for order reduction in vector-valued polynomial fits

tahandy:
When fitting vector-valued polynomial coefficients, the order for an individual component's fit may be _lower_ than the requested order. This can occur if the highest order coefficient (and any sequentially lower order coefficients) are identically 0 -- these coefficients will be pruned from the polynomial in `transformInput` during the application of a (potentially noop) shift at the end of fitting. In these cases, assembly of the vector-valued coefficients would result in an OOB exception, as one or more polynomials may not have the appropriate number of coefficients. In the fix, missing high-order coefficients are set to 0.0.

Note that in the case of _all_ polynomial orders resulting in a reduced order we still return a polynomial of the requested order, with the missing coefficients set to 0.0. Ideally it would return a polynomial that is the maximum of the individual coefficient polynomials, but I can't guarantee that someone, somewhere isn't directly accessing these coefficients and making the assumption that there is going to be `order+1` of them.

### DIFF
--- a/modules/c++/math.poly/include/math/poly/Fit.h
+++ b/modules/c++/math.poly/include/math/poly/Fit.h
@@ -331,14 +331,27 @@ inline math::poly::OneD< math::linear::VectorN< 3, double > > fit(
     const math::poly::OneD<double> fit1 = fit(xObs, yObs1, order);
     const math::poly::OneD<double> fit2 = fit(xObs, yObs2, order);
 
+    // There is a non-zero chance that one or more of the resulting polynomials
+    // are of a lower order than the requested order -- this results when the
+    // highest order coefficients are identically 0.0. As a result, care must be
+    // taken when combining the coefficients to avoid running outside of the
+    // actual number of coefficients in any particular fit.
+    // Note: There is a choice to be made here if /all/ polynomials have their
+    //       orders reduced. In an ideal world, this would reduce the order
+    //       of the output poly from 'order' to the maximum of the component
+    //       orders. However, this code has been in production for awhile,
+    //       and I cannot guarantee that someone, somewhere, isn't expecting
+    //       'order'+1 coefficients and attempting to extract/modify them.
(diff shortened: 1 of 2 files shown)


modules/c++/math.poly/unittests/test_llsq.cpp
```diff
@@ -237,4 +340,5 @@ int main(int, char**)
     TEST_CHECK(test1DPolyfitLarge);
     TEST_CHECK(test2DPolyfit);
     TEST_CHECK(test2DPolyfitLarge);
+    TEST_CHECK(testVectorValuedOrderChange);
```
asylvest: Pre-dates your update, but add a `return 0;` at the bottom of this

modules/c++/math.poly/include/math/poly/Fit.h
```diff
@@ -331,14 +331,27 @@ inline math::poly::OneD< math::linear::VectorN< 3, double > > fit(
...
+    //       and simply fill with zeros.
     math::poly::OneD< math::linear::VectorN< 3, double > > polyVector3 =
         math::poly::OneD< math::linear::VectorN< 3, double > >(order);
     for (size_t term = 0; term <= order; term++)
     {
         math::linear::VectorN< 3, double >& coeffs = polyVector3[term];
-        coeffs[0] = fit0[term];
-        coeffs[1] = fit1[term];
-        coeffs[2] = fit2[term];
+        coeffs[0] = (term <= fit0.order()) ? fit0[term] : 0.0;
```
asylvest: This is right on the border of the pedantic part of me wanting a separate inline function that does this so this code isn't cut and paste three times, but obviously the check is trivial and it's going to be more SLOC to put the new "helper" function in than to just do this so... ok, it's good as-is.

tahandy:
K, added requested changes.

PR #300, opened 2019-03-21
Update str::trim to properly handle non-ASCII characters on Windows

acwagen:


jmrandol:
We probably need to include <cwctype> at the top

acwagen:
@jmrandol I added wctype.h as an include.

jmrandol:
Oh, does including cwctype (C++ style vs. C style wctype.h) not work?

acwagen:
No, it doesn't, it can't find cwctype.h (on both Linux and Windows). The existing include was for ctype.h as well, not cctype.h.

jmrandol:
Hmm, usually when you include `#include <cwctype>` without the .h extension C++ will find the right header file

acwagen:
Should be good now!

PR #303, opened 2019-05-07
Update waf to work on Windows with Python3

JonathanMeans:
Diff:

    106c106
    <               except Exception as e:
    --- 
    >               except Exception,e:
    267c267
    <               except conf.errors.ConfigurationError as e:
    ---
    >               except conf.errors.ConfigurationError,e: 
    392c392
    <                                       except conf.errors.ConfigurationError as e:
    ---
    >                                       except conf.errors.ConfigurationError,e:
    701c701
    <                       os.write(fd,'\r\n'.join(i.replace('\\','\\\\')for i in cmd[1:]).encode())
    ---
    >                       os.write(fd,'\r\n'.join(i.replace('\\','\\\\')for i in cmd[1:]))


JonathanMeans:
If we have to do much more of this, would it make sense to just set up a repo for our own fork?

PR #304, opened 2019-05-24
Physical core count

tahandy:
Added `OS::getNumPhysicalCPUs()` to return the physical core count (as opposed to `OS::getNumCPUs()` which returns the logical core count).

__Unix implementation__: Scan for `thread_siblings_list` files under `/sys/devices/system/cpu/cpu*/topology/` and count the number of unique entries across files. 

__Windows implementation__: Throw `NotYetImplemented`. Not as simple as `getNumCPUs()`, and not needed at this time. A reference implementation for how to use `GetLogicalProcessorInformationEx` is at https://devblogs.microsoft.com/oldnewthing/?p=2823.

### DIFF
--- a/modules/c++/sys/include/sys/AbstractOS.h
+++ b/modules/c++/sys/include/sys/AbstractOS.h
@@ -215,7 +215,15 @@ public:
 
     virtual std::string getDSOSuffix() const = 0;
 
+    /*!
+     * Get the number of logical CPUs available (includes hyperthreading)
+     */
     virtual size_t getNumCPUs() const = 0;
+
+    /*!
+     * Get the number of physical CPUs available (excludes hyperthreading)
+     */
+    virtual size_t getNumPhysicalCPUs() const = 0;
 
     /*!
      *  Create a symlink, pathnames can be either absolute or relative
--- a/modules/c++/sys/include/sys/AbstractOS.h
+++ b/modules/c++/sys/include/sys/AbstractOS.h
@@ -216,12 +216,12 @@ public:
     virtual std::string getDSOSuffix() const = 0;
 
     /*!
-     * Get the number of logical CPUs available (includes hyperthreading)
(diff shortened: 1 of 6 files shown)


modules/c++/sys/include/sys/AbstractOS.h
```diff
@@ -215,8 +215,16 @@ class AbstractOS
 
     virtual std::string getDSOSuffix() const = 0;
 
+    /*!
+     * Get the number of logical CPUs available (includes hyperthreading)
```
asylvest: Use `\return` for these

modules/c++/sys/include/sys/OSWin32.h
```diff
@@ -181,12 +181,21 @@ class OSWin32 : public AbstractOS
...
     virtual void unsetEnv(const std::string& var);
 
+    /*!
+     * Get the number of logical CPUs available (includes hyperthreading)
+     */
     virtual size_t getNumCPUs() const;
 
+    /*!
+     * Get the number of physical CPUs available (excludes hyperthreading)
+     * TODO: Not yet implemented
```
asylvest: Use `\todo`

modules/c++/sys/source/OSWin32.cpp
```diff
@@ -281,6 +281,12 @@ size_t sys::OSWin32::getNumCPUs() const
     return info.dwNumberOfProcessors;
 }
 
+size_t sys::OSWin32::getNumPhysicalCPUs() const
+{
+    throw except::NotImplementedException(
```
asylvest: Put a TODO with the reference to the link you put in the PR so that we can look at that if we ever need to implement.  Here, use `// TODO` since not in Doxygen :o)

tahandy:
K, added proper doxygen and URL comment

PR #305, opened 2019-05-24
Affinity-based CPU reporting

tahandy:
Added affinity-based CPU count reporting to `OS`, available through `getNumCPUsAvailable` and `getNumPhysicalCPUsAvailable`. On Linux, these functions use `sched_getaffinity` and associated helpers to report total and physical CPU counts that may have been restricted by `numactl`/`taskset`. 

On Windows, these functions throw `NotImplementedException`.

### DIFF
--- a/modules/c++/sys/tests/CPUTest.cpp
+++ b/modules/c++/sys/tests/CPUTest.cpp
@@ -1,0 +1,51 @@
+/* =========================================================================
+ * This file is part of sys-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2019, MDA Information Systems LLC
+ *
+ * sys-c++ is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+#include <import/sys.h>
+
+using namespace sys;
+
+int main(int argc, char **argv)
+{
+    try
+    {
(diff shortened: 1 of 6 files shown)


modules/c++/sys/tests/CPUTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; If not,
+ * see <http://www.gnu.org/licenses/>.
+ *
+ */
+#include <import/sys.h>
+
+using namespace sys;
```
asylvest: Remove

modules/c++/sys/tests/CPUTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+int main(int argc, char **argv)
```
asylvest: Comment out argc and argv (or else VS will give annoying warnings

modules/c++/sys/tests/CPUTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+        std::cout << "Present number of CPUs: "
+                  << os.getNumCPUs() << std::endl;
+        std::cout << "Present number of physical CPUs: "
+                  << os.getNumPhysicalCPUs() << std::endl;
+        std::cout << "Available number of CPUs: "
+                  << os.getNumCPUsAvailable() << std::endl;
+        std::cout << "Available number of physical CPUs: "
+                  << os.getNumPhysicalCPUsAvailable() << std::endl;
+    }
+    catch (except::Throwable& t)
```
asylvest: Can be const

modules/c++/sys/tests/CPUTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+    {
+        std::cerr << "Caught throwable: " << t.toString() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+    catch (...)
+    {
+        std::cerr << "Caught unnamed exception" << std::endl;
```
asylvest: Should return 1 here.  Change the other exit to a return 1 (I assume this file was cut & paste from some other, lamer, test program)

tahandy: That would be correct...

tahandy:
K, changed CPUTest to not look like other tests in the sys module. ;-)

PR #306, opened 2019-05-29
CPU counting for large machines

tahandy:
Updates to CPU count functionality on UNIX if the number of CPUs available exceeds the capacity of the default size of `cpu_set_t`.

### DIFF
--- a/modules/c++/sys/source/OSUnix.cpp
+++ b/modules/c++/sys/source/OSUnix.cpp
@@ -133,9 +133,9 @@ std::set<std::string> get_unique_thread_siblings()
     return unique_ts;
 }
 
-struct CPUMask
+struct ScopedCPUMask
 {
-    CPUMask(int numCPUs)
+    ScopedCPUMask(int numCPUs)
     {
         mSize = CPU_ALLOC_SIZE(numCPUs);
         mMask = CPU_ALLOC(numCPUs);
@@ -148,7 +148,7 @@ struct CPUMask
         }
     }
 
-    ~CPUMask()
+    ~ScopedCPUMask()
     {
         if (mMask != NULL)
         {
@@ -170,7 +170,7 @@ public:
         // the constant CPU_SETSIZE.
         const size_t numOnlineCPUs = sys::OS().getNumCPUs();
         const int maxCPUs = std::max<int>(numOnlineCPUs, CPU_SETSIZE);
-        std::auto_ptr<CPUMask> cpuMask(new CPUMask(maxCPUs));
+        std::auto_ptr<ScopedCPUMask> cpuMask(new ScopedCPUMask(maxCPUs));
 
         if (sched_getaffinity(0, cpuMask->mSize, cpuMask->mMask) == -1)
         {
@@ -212,7 +212,7 @@ public:
     }
 
 private:
-    std::auto_ptr<const CPUMask> mCPUMask;
+    std::auto_ptr<const ScopedCPUMask> mCPUMask;
 };
 }
 
--- a/modules/c++/sys/source/OSUnix.cpp
+++ b/modules/c++/sys/source/OSUnix.cpp
@@ -33,6 +33,7 @@
 #include <set>
 #include <fstream>
 #include <sched.h>
+#include <errno.h>
 
 #if defined(__APPLE__)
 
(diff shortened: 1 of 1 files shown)


modules/c++/sys/source/OSUnix.cpp
```diff
@@ -132,13 +133,68 @@ std::set<std::string> get_unique_thread_siblings()
...
 
-void get_cpu_affinity(cpu_set_t* mask)
+class ScopedCPUAffinity
 {
-    if (-1 == sched_getaffinity(0, sizeof(cpu_set_t), mask))
+public:
+    ScopedCPUAffinity()
     {
-        throw except::Exception(Ctxt("Failed to get CPU affinity"));
+        int numOnlineCPUs = sysconf(_SC_NPROCESSORS_ONLN);
```
asylvest: Can be const

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -132,13 +133,68 @@ std::set<std::string> get_unique_thread_siblings()
...
+        if (numOnlineCPUs == -1)
+        {
+            throw except::Exception(Ctxt("Failed to get online CPU count"));
+        }
+
+        // The number of processors in the mask must be _at least_ the number of
+        // CPU's representable by the kernel's internal mask. This is given by
+        // the constant CPU_SETSIZE.
+        int maxCPUs = std::max(numOnlineCPUs, CPU_SETSIZE);
```
asylvest: Can be const.  Also, is CPU_SETSIZE just a `#define`?  If so, use `std::max<int>()` here just to make sure a compiler doesn't get confused and can't find a matching type or something dumb.

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -132,13 +133,68 @@ std::set<std::string> get_unique_thread_siblings()
...
-void get_cpu_affinity(cpu_set_t* mask)
+class ScopedCPUAffinity
 {
-    if (-1 == sched_getaffinity(0, sizeof(cpu_set_t), mask))
+public:
+    ScopedCPUAffinity()
     {
-        throw except::Exception(Ctxt("Failed to get CPU affinity"));
+        int numOnlineCPUs = sysconf(_SC_NPROCESSORS_ONLN);
+        if (numOnlineCPUs == -1)
```
asylvest: The current `OSUnix::getNumCPUs()` doesn't do this check (and it should).  Can you move that exception down there and then call the OS method here?  Then change type to `size_t`.

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -132,13 +133,68 @@ std::set<std::string> get_unique_thread_siblings()
...
+        }
+
+        // The number of processors in the mask must be _at least_ the number of
+        // CPU's representable by the kernel's internal mask. This is given by
+        // the constant CPU_SETSIZE.
+        int maxCPUs = std::max(numOnlineCPUs, CPU_SETSIZE);
+        mSize = CPU_ALLOC_SIZE(maxCPUs);
+        mMask = CPU_ALLOC(maxCPUs);
+
+        if (-1 == sched_getaffinity(0, mSize, mMask))
```
asylvest: Flip the order here (`sched_getaffinity() == -1`) because.... it just looks better that way

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -132,13 +133,68 @@ std::set<std::string> get_unique_thread_siblings()
...
+        {
+            throw except::Exception(Ctxt("Failed to get online CPU count"));
+        }
+
+        // The number of processors in the mask must be _at least_ the number of
+        // CPU's representable by the kernel's internal mask. This is given by
+        // the constant CPU_SETSIZE.
+        int maxCPUs = std::max(numOnlineCPUs, CPU_SETSIZE);
+        mSize = CPU_ALLOC_SIZE(maxCPUs);
+        mMask = CPU_ALLOC(maxCPUs);
```
asylvest: So this'll actually leak if the check below `sched_getaffinity()` throws because you'll never get into the destructor.  So you need an RAII class just dealing with this allocation/freeing.

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -132,13 +133,68 @@ std::set<std::string> get_unique_thread_siblings()
...
+        if (-1 == sched_getaffinity(0, mSize, mMask))
+        {
+            std::ostringstream msg;
+            msg << "Failed to get CPU affinity with"
+                << " CPU_SETSIZE=" << CPU_SETSIZE << ","
+                << " numOnlineCPUs=" << numOnlineCPUs << ","
+                << " alloc size=" << mSize << ".";
+            switch (errno)
+            {
+                case (EINVAL) :
```
asylvest: Don't indent the `case` inside `switch` statements (so move everything inside the body of the `switch` statement four spaces to the left)

modules/c++/sys/source/OSUnix.cpp
```diff
@@ -132,13 +133,68 @@ std::set<std::string> get_unique_thread_siblings()
...
+                    msg << " Affinity mask smaller than kernel mask size.";
+                    break;
+                case (EFAULT) :
+                    msg << " Invalid mask address.";
+                    break;
+                case (ESRCH) :
+                    msg << " PID for current process is invalid.";
+                    break;
```
asylvest: Add a default case

modules/c++/sys/source/OSUnix.cpp
asylvest: Don't put parens around the values (i.e. `case EINVAL:`)

modules/c++/sys/source/OSUnix.cpp
tahandy: Worth it to have a whole new class, or just a

``` 
if (mMask != NULL)
{
    CPU_FREE(mMask);
}
```

inside the failure if-statement?

asylvest: Let's have a whole new class to let RAII auto-magically handle everything for us

tahandy:
K, take a look.

asylvest:
Looks good

PR #307, opened 2019-06-03
Affinity thread pinning

tahandy:
1. Added waf configuration flag `--force-cpu-pinning` to, by default, turn on CPU pinning for `ThreadGroup` and the classes that use it. 

2. Restructured `CPUAffinityInitializer` and `CPUAffinityThreadInitializer` to have OS-specific implementations. GitHub seems to have gotten really confused on these diffs...

3. Wrapped generated `CPUAffinityThreadInitializer` objects created by `CPUAffinityInitializer` in `auto_ptr`. Note that there's a little trickiness to implement covariant `auto_ptr`s using name hiding (not `virtual`). 

4. `ScopedCPUMaskUnix` and `ScopedCPUAffinityUnix` are in their own file now (extracted from `OSUnix.h`). `ScopedCPUAffinityUnix` now derives from `ScopedCPUMaskUnix`.


modules/c++/mt/include/mt/CPUAffinityThreadInitializerLinux.h
```diff
@@ -0,0 +1,75 @@
...
+
+#if !defined(__APPLE_CC__)
+#if defined(__linux) || defined(__linux__)
+
+#include <memory>
+#include <sched.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#define gettid() syscall(SYS_gettid)
```
asylvest: This `#define` can go in the `.cpp` file to not pollute callers that include the header right?  And it looks like `gettid()`  is only called in one spot in the cpp, so do we need the `#define` at all vs. just calling this directly in the one spot?

tahandy: A relic of a bygone era...

PR #310, opened 2019-06-14
Fix writing to StringStream from Python

JonathanMeans:
About 9 months ago, I added #define SWIG_PYTHON_STRICT_BYTE_CHAR, to io.i, because there was a need to use StringStream with a Python bytes object. Unfortunately, this broke other things with Python3 that I didn't notice until recently.

This PR reverts that change, and adds a writeBytes method to the Python layer.

### DIFF
--- a/modules/python/io/unittests/test_string_conversions.py
+++ b/modules/python/io/unittests/test_string_conversions.py
@@ -2,12 +2,12 @@
 
 """
  * =========================================================================
- * This file is part of io-c++
+ * This file is part of io-python
  * =========================================================================
  *
  * (C) Copyright 2019, MDA Information Systems LLC
  *
- * io-c++ is free software; you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


modules/python/io/unittests/test_string_conversions.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+
+"""
+ * =========================================================================
+ * This file is part of logging-c++
```
asylvest: Fix module name

modules/python/io/unittests/test_string_conversions.py
```diff
@@ -0,0 +1,51 @@
...
+ * =========================================================================
+ *
+ * (C) Copyright 2019, MDA Information Systems LLC
+ *
+ * logging-c++ is free software; you can redistribute it and/or modify
```
asylvest: And here

PR #313, opened 2019-07-02
Make ThreadGroup default CPU pinning setting configurable at runtime

creffett:
* Added a static variable to ThreadGroup to contain the default pinning option (along with getter and setter)
* Added a method to ThreadGroup to get the current pinning option (inferred from whether `mAffinityInit` is set or a null pointer)
* Add unit tests to validate the new behavior
* Changed `mt_force_cpu_pinning` to `mt_default_cpu_pinning` in the mt wscript to reflect the change in behavior, likewise change the internal define from `MT_FORCE_PINNING` to `MT_DEFAULT_PINNING`

### DIFF
--- a/modules/c++/mt/include/mt/ThreadGroup.h
+++ b/modules/c++/mt/include/mt/ThreadGroup.h
@@ -88,12 +88,22 @@ public:
     /*!
      * Checks whether CPU pinning is set (i.e. whether mAffinityInit is NULL)
      *
-     * \returns true if CPU pinning is set, false otherwise
+     * \return true if CPU pinning is set, false otherwise
      */
     bool isPinToCPUEnabled() const;
 
+    /*!
+     * Gets the current default value for CPU pinning
+     *
+     * \return true if CPU pinning is enabled by default, false otherwise
+     */
     static bool getDefaultPinToCPU();
 
+    /*!
+     * Sets the default value for CPU pinning
+     *
+     * \param newDefault the new default value for CPU pinning
+     */
     static void setDefaultPinToCPU(const bool newDefault);
 
 private:
--- a/modules/c++/mt/include/mt/ThreadGroup.h
+++ b/modules/c++/mt/include/mt/ThreadGroup.h
@@ -57,15 +57,11 @@ public:
      * Constructor.
      * \param pinToCPU Optional flag specifying whether CPU pinning
      *                 should occur, using a CPUAffinityInitializer.
-     *                 If MT_FORCE_PINNING is defined, defaults to true
+     *                 If MT_DEFAULT_PINNING is defined, defaults to true
      *                 (enable affinity-based thread pinning).
      *                 Otherwise, defaults to false (no pinning).
      */
-#if defined(MT_FORCE_PINNING)
-    ThreadGroup(bool pinToCPU = true);
-#else
-    ThreadGroup(bool pinToCPU = false);
-#endif
+    ThreadGroup(bool pinToCPU = getDefaultPinToCPU());
 
     /*!
     *  Destructor. Attempts to join all threads.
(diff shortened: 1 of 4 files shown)


modules/c++/mt/include/mt/ThreadGroup.h
```diff
@@ -89,13 +85,31 @@ class ThreadGroup
...
     void joinAll();
 
+    /*!
+     * Checks whether CPU pinning is set (i.e. whether mAffinityInit is NULL)
+     *
+     * \returns true if CPU pinning is set, false otherwise
+     */
+    bool isPinToCPUEnabled() const;
+
+    static bool getDefaultPinToCPU();
```
asylvest: Add Doxygen to the new public methods

modules/c++/mt/include/mt/ThreadGroup.h
```diff
@@ -89,13 +85,31 @@ class ThreadGroup
      */
     void joinAll();
 
+    /*!
+     * Checks whether CPU pinning is set (i.e. whether mAffinityInit is NULL)
+     *
+     * \returns true if CPU pinning is set, false otherwise
```
asylvest: Use `\return` rather than `\returns` (for consistency with existing documentation)

modules/c++/mt/include/mt/ThreadGroup.h
```diff
@@ -89,13 +85,31 @@ class ThreadGroup
...
+     */
+    bool isPinToCPUEnabled() const;
+
+    static bool getDefaultPinToCPU();
+
+    static void setDefaultPinToCPU(const bool newDefault);
```
asylvest: I love `const` in general but don't use it when passing in variables by value into functions (reasoning being that at the caller level, they should know that the function can't change incoming variables passed by value in a way that affects them, so it doesn't improve interface readability... it does technically inform someone looking at the implementation of the method that the variable isn't going to be changed, but style-wise we're saying that the clutter of the extra `const` outweighs that benefit)

modules/c++/mt/include/mt/ThreadGroup.h
```diff
@@ -89,13 +85,31 @@ class ThreadGroup
...
     std::vector<mem::SharedPtr<sys::Thread> > mThreads;
     std::vector<except::Exception> mExceptions;
     sys::Mutex mMutex;
 
+    /*!
+     * The default setting for pinToCPU, used in the ThreadGroup constructor
+     * Can't set non-const static values in the header, so set at the top of
+     * the implementation
+     */
+    static bool defaultPinToCPU;
```
asylvest: For static or constant variables, go with underscore and all caps - `DEFAULT_PIN_TO_CPU`

modules/c++/mt/source/ThreadGroup.cpp
```diff
@@ -26,6 +26,14 @@
 
 namespace mt
 {
+
```
asylvest: Remove extra newline

asylvest:
This fundamentally looks sane to me... @tahandy any concerns?

modules/c++/mt/include/mt/ThreadGroup.h
creffett: Done. Might be worth checking documentation consistency; I see `\returns` in a few places in the existing docs.

modules/c++/mt/unittests/ThreadGroupTest.cpp
```diff
@@ -80,9 +80,36 @@ TEST_CASE(ThreadGroupTest)
...
+    ThreadGroup::setDefaultPinToCPU(false);
+    TEST_ASSERT_EQ(ThreadGroup::getDefaultPinToCPU(), false);
+    ThreadGroup threads3;
+    TEST_ASSERT_EQ(threads3.isPinToCPUEnabled(), false);
+}
+
 int main(int, char**)
 {
     TEST_CHECK(ThreadGroupTest);
+    TEST_CHECK(PinToCPUTest)
```
tahandy: Extra newline (you didn't add it, but let's get rid of it)

tahandy:
Format-wise, just that empty line in `ThreadGroupTest`.

I'm torn whether `isPinToCPUEnabled` is a bit of YAGNI or not. Do we need this downstream or is this just for having a unit test? I'd err toward keeping it, but if we don't intend on using it for anything... 🤷‍♂

modules/c++/mt/unittests/ThreadGroupTest.cpp
tahandy: Empty newline (you didn't add it, but let's remove it)

asylvest: Also, put a semicolon at the end of the line... which, granted, strictly speaking we don't actually need based on what that macro is... but it looks weird to end these lines without a semicolon and the rest of our unittests do

asylvest:
I mostly agree that we probably won't use `isPinToCPUEnabled()` outside of a unittest but it's a trivial method and it does let us actually unittest it so it makes sense to me to keep it

tahandy:
Sounds good.

PR #316, opened 2019-07-31
Fail configuration if trying to use too-old SWIG

JonathanMeans:
Sometimes people try to build with SWIG with some ancient 1.x version, and the error messages aren't great when it inevitably fails. 

I just set it by major version, so if you want to use, say, 3.0.7 for your project you can, but it'll catch a setup that just doesn't make sense. 

Then I added at the top level-wscript that only gets called if you're building CODA-OSS itself, a check for the current version. Presumably if you're building with SWIG, that means you plan to change the Python bindings and check it in, so you should definitely be using the right version for that.

### DIFF
--- a/build/swig.py
+++ b/build/swig.py
@@ -195,7 +195,7 @@ def options(opt):
     opt.add_option('--disable-swig', action='store_false', dest='swig',
                    help='Disable swig')
     opt.add_option('--swig-version', action='store', dest='swigver',
-                   default=None, help='Specify the minimum swig version')
+                   default='3.0.0', help='Specify the minimum swig version')
     opt.add_option('--require-swig', action='store_true', dest='require_swig',
                    help='Require swig (configure option)', default=False)
 
--- a/wscript
+++ b/wscript
@@ -18,6 +18,7 @@ def options(opt):
     opt.recurse(DIRS)
 
 def configure(conf):
+    conf.options.swigver = '3.0.12'
     conf.load(TOOLS, tooldir='./build/')
     conf.recurse(DIRS)
 


asylvest:
This makes sense to me.  So if you configure with an old version of Swig, what error message do you get with this change?

JonathanMeans:
```
Checking for program swig                               : /opt/swig/3.0.10/bin/swig
Traceback (most recent call last):
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Scripting.py", line 97, in waf_entry_point
    run_commands()
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Scripting.py", line 153, in run_commands
    ctx=run_command(cmd_name)
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Scripting.py", line 146, in run_command
    ctx.execute()
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Configure.py", line 128, in execute
    super(ConfigurationContext,self).execute()
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Context.py", line 84, in execute
    self.recurse([os.path.dirname(g_module.root_path)])
  File "/data1/u/jmeans/coda-oss/build/build.py", line 75, in recurse
    super(CPPContext, self).recurse(dirsWithWscripts)
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Context.py", line 125, in recurse
    user_function(self)
  File "/data1/u/jmeans/coda-oss/wscript", line 22, in configure
    conf.load(TOOLS, tooldir='./build/')
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Configure.py", line 193, in load
    if type(func)is type(Utils.readf):func(self)
  File "./build/swig.py", line 209, in configure
    conf.check_swig_version(minver=swigver)
  File "/data1/u/jmeans/coda-oss/build/.waf-1.7.14-872d6f34b84e4daf914a1e2b7fe333fd/waflib/Configure.py", line 224, in fun
    return f(*k,**kw)
  File "./build/swig.py", line 185, in check_swig_version
    % (minver_str, swigver_full))
RuntimeError: Minimum SWIG verion required: 3.0.12. Version found: 3.0.10

```

asylvest:
OK cool - that is a helpful message.

PR #321, opened 2019-08-16
Added TEST_CASE_ARGS and TEST_CHECK_ARGS for c++

vkrishna1997:


JonathanMeans:
Make sure you tag the reviewers/assignees/other stuff in the column on the right. (Adam's the one who actually has the permission to merge it all in.)

Has anyone shown you the [style guide](http://mi-gitlab.mdaus.corp/code/coda/wikis/cpp-coding-standards)? (Do you have access to the internal Gitlab?) If you can't get to it, the super-condensed version would be: indentation is 4 spaces, curly braces get their own line, and a single space after `if`/`for`/`while`.

asylvest:
So usually we run unittests with no command-line arguments (though technically nothing in `TestCase.h` requires that)... what's the use case for adding the check that these macros implement?

vkrishna1997:
@JonathanMeans thanks for the info. I've been following your style for the six implementation but the style guide is helpful!
@asylvest Gotcha. Because we don't have any XML docs for CPHD yet, I believe Jonathan had made his own XML string in one of the tests that contains an XML document for testing parsing. I wanted to be able to test specific variations of the XML (for ex: XML with different values or types and invalid values). This would require multiple extremely long strings for different values so I wanted to use fileInputStream to read in an XML file.

JonathanMeans:
I agree that it makes sense to have some standalone test to make sure we can read in CPHD XML. I don't think the TestCase macros are a good tool for that, for two reasons:

 - There are Jenkins jobs for the GitHub repos that automatically go into the installed `unittests` directories, run everything in there that they can find with no arguments, and fail on a non-zero return code. I'm worried this will encourage people to write code that violates that convention.

 - The unittest macros are convenient for when you need to make a lot of checks and easily pinpoint what went wrong. If you're taking in arbitrary xml files, I don't think you could say very much beyond that it built a valid object and maybe round-tripped okay, so I don't see that the macros would be helping much. 

I don't have any great ideas for how to solve your actual problem, so I'll have to defer to Adam there. Maybe it's sufficient to just make sure the XML can round-trip? (i.e. you get the same string if you read into memory and write it back out) (FWIW, I believe the CPHD testing code I left you with was an "experiment" for me that I was starting to feel was too tedious to be valuable, IIRC)

PR #327, opened 2019-10-08
Add global config header

JonathanMeans:
Asking for two pairs of eyes here since this is a little big.

First change is adding a new `config` module that just generates the config header and doesn't depend on anything else. (I'm working to make the C++ layer of NITRO available as a shared library. Which means making CODA-OSS available as a shared library. Which means adding macros to conditionally expand to `__declspec(dllexport)` everywhere. I therefore need an independent step to ensure that everyone has access to the header defining the macro.)

While I was in there, I found a switch to let waf automatically run your unit tests, so I went ahead and exposed that. The reason I moved `test_symlink.cpp` from `unittests` to `tests` is because some Windows configurations (e.g. mine) don't have permissions to create symlinks, and I don't think that's a good reason for an automatic test to fail.


modules/c++/config/wscript
```diff
@@ -0,0 +1,84 @@
...
+                v = int(v)
+            except:
+                v = v.strip()
+                if v == 'True':
+                    v = True
+                elif v == 'False':
+                    v = False
+            conf.define(k.upper(), v)
+        conf.check_cxx(fragment = cpp11_str, execute=1, msg='Checking for C++11 support', define_name='__CODA_CPP11', mandatory=False)
+        conf.define('CODA_EXPORT', conf.env['declspec_decoration'], quote=False)
```
jmrandol: How does declspec_decoration get into the environment?

JonathanMeans: https://github.com/mdaus/coda-oss/blob/master/build/build.py#L1301

wscript
```diff
@@ -24,6 +24,7 @@ def configure(conf):
 
 def build(bld):
     bld.recurse(DIRS)
+    enableWafUnitTests(bld)
```
clydestanfield: Does this always run all unittests after building or is there a flag I set at configure time? How does this compare with the existing "--unittests" flag (I honestly don't know how the existing flag works, I just know that Jenkins sets it)?

JonathanMeans: The previous behavior for the `--unittests` flag was to mark unit tests as tests when passed. As far as I can tell, this causes waf to internal run the tests, but not tell you if they passed or not. I've just removed it.

The new behavior, if you build the first time, it will automatically run all of the tests and return non-zero on failure. Then if you build again, it will only re-run the tests that needed to recompile. (Unless you pass `--alltests`). As far as I know, projects that use coda-oss jump over this top-level wscript, so their behavior will be unaffected.

PR #329, opened 2019-10-16
Only recurse into directories that exist

JonathanMeans:


JonathanMeans:
Hang on, I need to regenerate the Python bindings too

JonathanMeans:
Ignore that last note, Python bindings are up to date

PR #331, opened 2019-10-21
Fix incorrect checking in unittest

JonathanMeans:
This test was intermittently failing. Writing into erased segments was causing it to overwrite the expected data.

Is this correct? I'm not sure of the intended usage of this class, so possibly the original behavior was the desired operation and should have passed. 

The changes in ScratchBuffer.cpp were problems flagged by cppcheck while debugging.

### DIFF
--- a/modules/c++/mem/source/ScratchMemory.cpp
+++ b/modules/c++/mem/source/ScratchMemory.cpp
@@ -60,6 +60,10 @@ ScratchMemory::Segment::Segment(size_t numBytes,
 void ScratchMemory::release(const std::string& key)
 {
     std::map<std::string, Segment>::const_iterator iterSeg = mSegments.find(key);
+    if (iterSeg == mSegments.end())
+    {
+        throw except::Exception(Ctxt("Key " + key + " does not exist"));
+    }
     mReleasedKeys.insert(key);
 
     if (mKeyOrder.back() == key)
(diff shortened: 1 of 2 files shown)


modules/c++/mem/source/ScratchMemory.cpp
```diff
@@ -75,9 +75,9 @@ void ScratchMemory::release(const std::string& key)
                                                                mKeyOrder.end(),
                                                                key);
         std::vector<std::string>::iterator nextKeyIter = mKeyOrder.erase(keyIter);
+        const std::string nextKey = *nextKeyIter;
         mKeyOrder.push_back(key);
 
-        const std::string nextKey = *nextKeyIter;
```
kdwilhelm: There are references to `nextKeyIter` later in this function, which may be invalidated when `mKeyOrder.push_back(key)` reallocates. I suggest replacing all of the `*nextKeyIter` after the `push_back` with `nextKey`.

JonathanMeans: Good catch! There was a spot below where it was actually relying on the iterator in a messy way, so I put the `push_back` operation at the start. Then when it searches the vector it should still find the original element.

I'm not sure how that behaves if the key was not in the vector originally. I didn't see any checks for if it wasn't found. Should that be able to happen?

kdwilhelm: It looks to me like if the key is in `mSegments`, it must also be in `mKeyOrder`.

I think the first line in the `release` method `mSegments.find(key)` should probably either return or throw if the key is not found.

JonathanMeans: How's this?

kdwilhelm: That looks good to me.

PR #334, opened 2019-11-05
Don't use the -m64 flag with GCC/ICC, and add support for OpenMP

wgvisots:
Ideally I would have removed the 64bit flag further up, where there is a TODO to remove it.  But I don't have a way to test Windows/Sun/Darwin.  So I am just replacing it with an empty string for gcc/icc.

The `-m64` flag forces the compiler to do pedantic IEEE-784 math, which does a bunch of really slow and unnecessary testing with every operation.  We never check the results of those checks, so we shouldn't do them.  In addition to this, we should add some `-march=...` flag

The OpenMP flag allows for `#pragma omp ...` statements.  Leaving the flag enabled by default doesn't do anything if there is no OpenMP code to run.  Disabling it lets the compiler single-thread the OpenMP-parallelized code.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -792,6 +792,9 @@ def options(opt):
                    help='Specify a prebuilt modules config file (created from dumpconfig)')
     opt.add_option('--disable-swig-silent-leak', action='store_false', dest='swig_silent_leak',
                    default=True, help='Allow swig to print memory leaks it detects')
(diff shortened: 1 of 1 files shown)


creffett:
d'oh, forgot about this, sorry. Will try to review this tomorrow.

creffett:
@wgvisots I don't see anything obviously wrong, but I've never heard of this -m64 behavior you describe - can you link me some documentation on that?

PR #338, opened 2019-11-25
Fix incorrect range check. Seconds are floating-point values,

JonathanMeans:
so, e.g., 59.23 should be valid.


vkrishna1997:
The fixes look good to me.

PR #343, opened 2020-02-11
Add a .clang-format file based on GitLab style guide

JonathanMeans:
This style file produces an empty/minimal diff on most files I ran it on. Most of the changes it makes/suggests are for issues that would have been a violation anyway. 

My intention is require this for "my" projects (SIX, NITRO). 

For wider CODA, I worry that this will be rather useless without widespread adoption, and I'm not sure how to make that happen.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,10 +1,10 @@
 BasedOnStyle: Google
 AccessModifierOffset: '0'
 AlignAfterOpenBracket: Align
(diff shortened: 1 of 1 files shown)


jmrandol:
Looks fine to me.  I think for CODA you just need to do module by module as they're touched, then it can be enforced in the builds.

PR #354, opened 2020-04-24
Added circular buffer

chvink:


tahandy:
One last thing, you'll want to drop the license at the top of these files. I've updated it to be relevant to 2020:

```
/* =========================================================================
 * This file is part of mem-c++
 * =========================================================================
 *
 * (C) Copyright 2004 - 2020, Radiant Geospatial Solutions
 *
 * mem-c++ is free software; you can redistribute it and/or modify
 * it under the terms of the GNU Lesser General Public License as published by
 * the Free Software Foundation; either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public
 * License along with this program; If not,
 * see <http://www.gnu.org/licenses/>.
 *
 */
```

chvink:
Should be all set

jmrandol:
Any reason why this isn't merged?

chvink:
I'm not sure if anyone is going to be able to get around to using this in the spot it was originally intended any time soon, but I think it can be merged anyway on the off chance someone wants a circular buffer for something.

PR #359, opened 2020-06-30
Added ability to ignore unknown arguments in the ArgumentParser.

fourleggedfriends:


modules/c++/cli/source/ArgumentParser.cpp
```diff
@@ -437,24 +472,34 @@ cli::Results* cli::ArgumentParser::parse(const std::vector<std::string>& args)
...
                 else
                 {
-                    throw except::Exception(Ctxt(FmtX("Invalid option: [%s]",
-                                                      argStr.c_str())));
+                    if (mIgnoreUnknownArguments)
+                    {
+                        mIgnoreUnknownOStream
+                                << "Unknown arg: " << argStr.c_str()
+                                << std::endl;
+                        break;
```
kdwilhelm, line 480: This break will cause all arguments after this one to be ignored. Changing it to `continue` will allow it to process the remaining arguments.

There are also some more complicated cases to consider, like if a user tries to pass an argument which does not exist with a following value (something like `--filename C:/Data/File.txt`).

modules/c++/cli/unittests/test_cli.cpp
```diff
@@ -175,6 +177,53 @@ TEST_CASE(testRequired)
...
 
+TEST_CASE(testUnknownArgumentsOptions)
+{
+    cli::ArgumentParser parser(true, std::cerr);
+    parser.setProgram("tester");
+    parser.addArgument("-v --verbose", "Toggle verbose", cli::STORE_TRUE);
+    parser.addArgument("-x --extra", "Extra options", cli::SUB_OPTIONS);
+
+    // Use a flag that is incorrect
+    std::auto_ptr<cli::Results> results(parser.parse(str::split("-z", " ")));
```
kdwilhelm, line 188: `std::auto_ptr` is deprecated as of c++11 and removed in c++17, so this should be changed to `std::unique_ptr` to avoid breaking when we update to c++17.

fourleggedfriends:
Thanks for the helpful comments!

jmrandol (approved):

jmrandol:
Forgot about this ticket, but this looks good to merge to me

PR #363, opened 2020-09-01
fix g++ on WSL compiler warnings

JDanielSmith:
fix g++ on WSL compiler warnings


modules/c++/tiff/source/Common.cpp
```diff
@@ -54,7 +54,8 @@ sys::Uint64_T tiff::combine(sys::Uint32_T numerator,
 void tiff::split(sys::Uint64_T value, sys::Uint32_T &numerator,
                  sys::Uint32_T &denominator)
 {
-    numerator = ((sys::Uint32_T *)&value)[0];
-    denominator = ((sys::Uint32_T *)&value)[1];
+    const sys::Uint32_T* pValue = reinterpret_cast<sys::Uint32_T*>(&value);
+    numerator = pValue[0];
+    denominator = pValue[1];
```
kdwilhelm, line 59: Casting sys::Uint64_T* to sys::Uint32_T* and dereferencing the result violates the strict aliasing rule and is undefined behavior. It tends to work for some compiler/option combinations (e.g. gcc with -fno-strict-aliasing), but it is not standard.

A standard compliant way to do this (until we can use C++20 bit_cast) is
```
const auto* buf = reinterpret_cast<sys::ubyte*>(&value); // reinterpret casting to char/unsigned char is allowed by the standard
memcpy(&numerator, buf, sizeof(sys::Uint32_T));
memcpy(&denominator, buf + sizeof(sys::Uint32_T), sizeof(sys::Uint32_T));
```

These give some details on the aliasing rules in C++. There are probably many other places in coda-oss with similar issues.
https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8
https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing

modules/c++/tiff/include/tiff/Header.h
```diff
@@ -57,14 +57,25 @@ class Header : public io::Serializable
         mId(id), mIFDOffset(ifdOffset)
     {
         bool isBigEndian = sys::isBigEndianSystem();
+        // compilers are now very aggressive about checking propoer strncpy() usage
```
kdwilhelm, line 60: Typo in "proper", though I'm not sure this needs to be included since all the `strncpy` were removed.

JDanielSmith: 👍 fixed (and added a few more notes)

modules/c++/tiff/source/Common.cpp
JDanielSmith: Good call! I submitted an update

modules/c++/tiff/source/Common.cpp
```diff
@@ -39,22 +41,33 @@ std::string tiff::RationalPrintStrategy::toString(const sys::Uint32_T data)
...
 void tiff::split(sys::Uint64_T value, sys::Uint32_T &numerator,
                  sys::Uint32_T &denominator)
 {
-    numerator = ((sys::Uint32_T*)&value)[0];
-    denominator = ((sys::Uint32_T*)&value)[1];
+    // Using casts generate a warning: dereferencing type-punned pointer will break strict-aliasing rules[-Wstrict-aliasing]
+    // Do the reverse of combine() and use memcpy().
+    //numerator = ((sys::Uint32_T*)&value)[0];
+    //denominator = ((sys::Uint32_T*)&value)[1];
+    auto ptr = reinterpret_cast<const std::byte*>(&value);  // reinterpret_cast<> to std::byte is allowed by the standard
```
kdwilhelm, line 69: I would prefer to use `sys::ubyte*` (which is just a typedef of `unsigned char*`) here, or alternatively `unsigned char*` directly if `sys::ubyte*` gives a warning. I don't think defining `std::byte` for C++ versions < 17 ourselves is technically allowed (https://en.cppreference.com/w/cpp/language/extending_std).

Also, the GSL library has some more complicated conditions for testing whether `std::byte` is available:
https://github.com/beinhaerter/GSL/blob/128683c204616fcd441f1643cdcaf33a65a0392c/include/gsl/gsl_byte#L60-L64. I think it's better to avoid until coda-oss requires C++17.

JDanielSmith: OK ... done.

kdwilhelm (approved):
Looks good to me.

PR #372, opened 2020-10-20
Develop/write utf 8 strings

JDanielSmith:


JDanielSmith:
See #371  for details.

PR #462, opened 2021-02-02
make getValue() easier to use

JDanielSmith:
`getValue()` will now optionally `throw` instead of returning a `bool`.


modules/c++/xml.lite/include/xml/lite/Attributes.h
```diff
@@ -381,22 +386,27 @@ class Attributes
  * \param result The value after calling str::toType(), if found
  * \return If an attribute with the key is found or not
  */
-template <typename T, typename K, typename ToType>
-inline bool getValue_(const Attributes& attributes, const K& key, T& result,
-    ToType toType)
+template <typename K, typename ToType>
+inline auto castValue_(const Attributes& attributes, const K& key, ToType toType)
+  -> decltype(toType(std::string()))
```
dstarinshak, line 391: @JDanielSmith, this breaks the SWIG bindings. If you configure in a mode that regenerates the bindings (`DENABLE_SWIG=ON` for cmake, `--enable-swig --require-swig` for waf), this line generates a syntax error. SWIG currently does not support complex/compound `decltype` expressions involving an internal function. It only supports expressions like `decltype(x)`.

JDanielSmith: I just added `#ifndef SWIG` around this code in the [feature/xml.lite_tweaks](https://github.com/mdaus/coda-oss/tree/feature/xml.lite_tweaks) branch; does that fix the problem?

dstarinshak: Yeah, that appears to fix the build. I can import the SWIG-generated Python module, but I haven't verified if this impacts any functionality. What impact do you anticipate this having? Seems to me that commenting out a bunch of template code in a low-level module like this would have a big effect.

Also note that no released version of swig currently supports complex `decltype`'s. It's not just that we're using an older version. The release notes for the latest version (4.0.2) still mention this limitation.

JDanielSmith: It's brand new code, so it wouldn't be used in the SWIG wrappers; code that SWIG never sees (i.e., **cpp** files) can freely use.

PR #464, opened 2021-02-08
The (old) version of SWIG we're using doesn't like certain C++11 features.

JDanielSmith:
The (old) version of SWIG we're using doesn't like certain C++11 features.  Since we don't need such code in the SWIG wrappers, it's easy enough to "comment-out" with `#ifndef SWIG`.


dstarinshak:
Moving this comment here for posterity:

Yeah, that appears to fix the build. I can import the SWIG-generated Python module, but I haven't verified if this impacts any functionality. What impact do you anticipate this having? Seems to me that commenting out a bunch of template code in a low-level module like this would have a big effect.

JDanielSmith:
It's brand new code, so it wouldn't be used in the SWIG wrappers; code that SWIG never sees (i.e., **cpp** files) can freely use.

dstarinshak:
I'm still concerned there may be unintended side effects. Forgive me if this is a little meandering, but here's my thought process:

CODA has three configurations when it comes to python modules:
1. Do not install - occurs when python is disabled or if the configure step can't find python with sufficient version
2. Install generated bindings - compile the *_wrap.cxx files checked into the repo, create a *.so, and install the *.py wrapper
3. Generate new bindings - SWIG is invoked on the *.i files to generate a *_wrap.cpp and *.py file. Then it follows configuration 2

The difference between config 2 and config 3 is whether or not SWIG was passed at configure time. From what I can tell, config 3 will set the `SWIG` macro variable that's hiding the offending code.

The behavior I'm concerned about is what happens if I build with config 3 and get a shared object for the `xml_lite` bindings. Then I try to use them (or some other module's bindings that import `xml_lite`) and run into an undefined symbol somewhere because a .cpp file never used the code that was `#ifndef`'d away. The shared libraries for the Python modules will build fine and expect any undefined symbols to be present in .o/.a files in the install directory. The Python interpreter will try to fill in the symbols at import time and will throw if it can't find time.

JDanielSmith:
The newly added "problem code" (i.e., "advanced" C++11) is all `template`s and/or `inline`, there's no way for something else to compile but not have the implementation.

PR #493, opened 2021-04-12
Compiler optimizations

adam-beauchamp:
Pulling in some Konradian optimization tools of global utility

### DIFF
--- a/.github/workflows/build_unittest.yml
+++ b/.github/workflows/build_unittest.yml
@@ -78,6 +78,7 @@ jobs:
       run: |
         env
         which python
+        pip install numpy
         mkdir installLinux-Github
         python waf configure --prefix=$(realpath installLinux-Github)
     - name: build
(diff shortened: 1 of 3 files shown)


JDanielSmith:
I've made some "fairly recent" tweaks to use a compiler-version check rather than WAF configuration to figure this out; we can easily get both MSVC and GCC versions, which are the only? two compilers I think we care about.

Also, I've tried to be rigorous about prefixing any *new* macros with `CODA_OSS`, and having a `_` suffix for internal-use-only ones.

adam-beauchamp:
@JDanielSmith, it looks like `[develop/compiler-optimizations](https://github.com/mdaus/coda-oss/tree/develop/compiler-optimizations)`  are pretty close to each other;

Should we merge this branch into yours, and then yours into master? Let me know how you want to swing it...

PR #510, opened 2021-07-13
Fix wscript to successfully build test_num_elements

mbraswell:
There was a missing required library in this wscript which caused some tests to fail to build. This fixes the issue.


adam-beauchamp (commented):
looks sane

PR #511, opened 2021-07-16
Fix for numerical precision issues surrounding horizontal lines

adam-beauchamp:
Migrating a subtree patch to oss that applies the same principle to horizontal calculations as is applied to vertical ones


adam-beauchamp (commented):
LGTM

PR #515, opened 2021-07-29
Security and conan updates

kdwilhelm:
- The previous libjpeg and zlib versions were affected by multiple CVEs and have been updated to the current latest versions. 
- Most of the "drivers" were updated to support packaging with conan. Recipes for each one are exported to the local conan cache when the coda-oss recipe is exported. They can then be built automatically with coda-oss by passing the `--build outdated` option to conan when running `conan create` or `conan install`.
- Several warnings were fixed or disabled for MSVC 2017 and earlier (needed since the /WX (warnings as errors) flag was recently added).


JDanielSmith (approved):

modules/c++/gsl/include/gsl/Gsl_.h
```diff
@@ -66,12 +66,20 @@ namespace Gsl
         {
             return (static_cast<U>(t) != u) ? narrow_throw_exception(t) : t;
         }
+
+        #if defined(_MSC_VER) && _MSC_VER <= 1900
```
adam-beauchamp, line 70: @kdwilhelm, `modules/c++/config/include/config/compiler_extensions.h` now has some cross-platform macros for doing this sort of thing  look for `CODA_OSS_disable_warning_push` in the file

modules/c++/sys/tests/ThreadTest4.cpp
```diff
@@ -71,9 +70,12 @@ class Putter : public sys::Runnable
 public:
     Putter(sys::Mutex *by,int *val, int n) : theVal(val), syncBy(by), id(n)
     {}
-    virtual ~Putter()
-    {}
+    virtual ~Putter() = default;
 
+    #if _MSC_VER
```
adam-beauchamp, line 75:
```c++
CODA_OSS_disable_warning_push
CODA_OSS_DISABLE_UNREACHABLE_CODE
```

modules/c++/sys/tests/ThreadTest4.cpp
```diff
@@ -92,6 +94,10 @@ class Putter : public sys::Runnable
         }
 
     }
+    #if _MSC_VER
```
adam-beauchamp, line 97:
```c++
CODA_OSS_disable_warning_pop
```

modules/drivers/xml/xerces/CMakeLists.txt
```diff
@@ -1,306 +1,313 @@
-set(TARGET_NAME xerces-c)
```
adam-beauchamp, line 1: Looks like line endings got flipped on this file, can you flip them back

adam-beauchamp (changes requested):
@kdwilhelm, I added a few features this week that are directly relevant to your mods, hence throwing in my $0.02

Over to @jmrandol for the conan stuff (and anything else he cares to look at)

modules/drivers/xml/xerces/CMakeLists.txt
kdwilhelm: Line endings in this file were previously flipped from LF to CRLF in bc5552cd209eb937da82dc433aafed9eb89be1ca. This change restores the LF line endings. You can hide the whitespace changes in the Github diff: https://github.blog/2018-05-01-ignore-white-space-in-code-review/

I also added a .gitattributes file marking CMakeLists.txt (as well as for *.in, *.cmake, and conanfile.py) as `text` files, which turns on text line ending normalization in git (https://git-scm.com/docs/gitattributes). This sets git to use LF endings internally for these files, which should prevent this from happening again.

By default, git uses LF line endings internally for text files but converts to CRLF on checkout for Windows. I set an additional option in .gitattributes to force LF on checkout for all platforms for those 4 file types (`eol=lf`). This is needed to support conan recipe revisions. When a conan recipe is exported, the conanfile.py and any files listed in `exports_sources` are hashed to create a recipe revision id. Line ending differences change the hash result, which would cause the recipe ID to change if you upload a package built on Windows to a remote (such as an Artifactory server) and subsequently upload a package built on Linux to the same remote. These recipe revision ID changes due to whitespace would require an unnecessary rebuild of any packages built with the previous recipe revision ID.

JDanielSmith:
Looks good to me! 👍🏻👍🏻

I seem to recall trying “platform normalization” before … it didn’t work very well with WSL.  Checking out on Windows (CRLF) and then accessing those same files via WSL (/mnt/c/Users/…/source/repos/…) was “wrong.”

The advice I’ve been reading is to do everything *nix-style (just LF) as all the Windows tools (including Notepad) can handle that.

   Dan

modules/drivers/xml/xerces/CMakeLists.txt
adam-beauchamp: 👍

PR #516, opened 2021-07-29
This Fixes 3 compiler warnings and enables building with cpp17 on Windows.

mbraswell:


JDanielSmith (commented):
You probably want to put the #pragma warning(disable) inside a _MSC_VER #ifdef so that you don't get warnings with G++.

build/build.py
```diff
@@ -1053,6 +1053,11 @@ def configureCompilerOptions(self):
                   '_LARGEFILE_SOURCE WIN32 _USE_MATH_DEFINES NOMINMAX WIN32_LEAN_AND_MEAN'.split()
         flags = '/UUNICODE /U_UNICODE /EHs /GR'.split()
 
+        #If building with cpp17 add flags/defines to enable auto_ptr
+        if Options.options.enablecpp17:
+            defines.append('_HAS_AUTO_PTR_ETC=1')
+            flags.append('/std:c++17')
+
         self.env.append_value('DEFINES', defines)
```
JDanielSmith, line 1061: `std::auto_ptr` shouldn't be needed in C++17, use `mem::auto_ptr` instead.

JDanielSmith (approved):

adam-beauchamp:
@mbraswell, can you rebase against master and then repush your branche ?

We just added some macros to modules/c++/config/include/config/compiler_extensions.h that you can leverage to make the `#pragma` stuff work in a cross-platform way

build/build.py
adam-beauchamp: @JDanielSmith, can we just have the code emit some friendly advice about how to update to newer pointer types here, while still allowing the extra `_HAS_AUTO_PTR_ETC=1` ? 

I could see some older codebases preferring that (which I'm guessing is the driver for this change), since without this change, as soon as you try to play around w/ c++17, you get smacked in the face with build failures.

JDanielSmith: C++17 has **removed** `std::auto_ptr`, it has been deprecated since C++11.  If you want to "play around" w/C++17, **change your code**; `mem::auto_ptr` can help as it is still `std::auto_ptr` pre C++17.

All of **coda-oss**, **nitro** and **six-library** are "clean" in this regard; no special compiler flag needed to bring `std::auto_ptr` into C++17.  If *projects* want to do that, nothing stopping them; but not in **coda-oss**.

mbraswell: Okay I will remove that, but I will point out that std::auto_ptr is being enabled in the gcc flags with -Wno-deprecated-declarations, so if it is a problem to have removed features enabled, that flag should be removed as well.

JDanielSmith: Personally, I would have done all of this C++17 stuff in the **main** branch (not **master**) which was already C++17-enabled and does not support anything earlier.

mbraswell: The main branch has that compiler flag as well.

JDanielSmith (approved):

modules/c++/dbi/source/DatabaseClientFactory.cpp
```diff
@@ -24,7 +24,9 @@
 #include "dbi/MySQLConnection.h"
 #include "dbi/PgSQLConnection.h"
 #include "dbi/OracleConnection.h"
-
+#include "config/compiler_extensions.h"
+CODA_OSS_disable_warning_push
+CODA_OSS_disable_warning(4702)
```
adam-beauchamp, line 29: To make this work properly, you'll need to add something like the following to `modules/c++/config/include/config/compiler_extensions.h`
```c++
// in the msvc section...
#define CODA_OSS_DISABLE_UNREACHABLE_CODE CODA_OSS_disable_warning(4702)
// in the gcc/clang section and the #else ...
#define CODA_OSS_DISABLE_UNREACHABLE_CODE
```
and then use `CODA_OSS_DISABLE_UNREACHABLE_CODE` in this file

adam-beauchamp (changes requested):
Tweak your usage of the defines and the linux build will be healed

adam-beauchamp (approved):

PR #517, opened 2021-07-29
Macros to narrowly target disabling warnings

adam-beauchamp:
- [x] Macros for turning off warnings in a crossplatform way (https://www.fluentcpp.com/2019/08/30/how-to-disable-a-warning-in-cpp/)
- [x] Fix windows warning 4551 ('function call missing argument list') on one of the numpyutils tests (in the following code)
       ` CODA_OSS_mark_symbol_unused(_import_array);`
    - [x] Verified that no function call is actually emitted by MSVC when a function reference without an argument list appears in the code (like above) with the following demo program
```c++
// Compile with:
// cl.exe /EHsc foo.cpp
#include <iostream>

void
foo() {
   std::cout << "error if this shows up!" << std::endl;
}

int
main() {
    std::cout << "before" <<std::endl;
    foo;
    std::cout << "after" <<std::endl;
    return 0;
}
```

### DIFF
--- a/modules/c++/config/include/config/compiler_extensions.h
+++ b/modules/c++/config/include/config/compiler_extensions.h
@@ -86,11 +86,40 @@
     #endif
 #endif  // _MM256_EXTRACTF
 
+
+#ifndef CODA_OSS_HAVE_DISABLE_WARNINGS
+    #define CODA_OSS_HAVE_DISABLE_WARNINGS 1
+
+    // Adapted from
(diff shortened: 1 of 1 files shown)


JDanielSmith (approved):

JDanielSmith (approved):

adam-beauchamp:
Thanks @JDanielSmith !

PR #518, opened 2021-07-30
Bugfix/python3.9 related

adam-beauchamp:
- numpy's includes cause no end of trouble with the  `modules/c++/numpyutils/unittests/test_num_elements.cpp` test
    - I built locally against a 3.9.x version of python with MSVC, which manifested the error (not present in our default build against python 3.7)
    - I added some macros to allow turning off warnings when `#include`-ing a header or set of headers
      - Inspiration for the change came from this MS blogpost: https://devblogs.microsoft.com/cppblog/broken-warnings-theory/#warning-level-for-external-headers

### DIFF
--- a/modules/c++/config/include/config/compiler_extensions.h
+++ b/modules/c++/config/include/config/compiler_extensions.h
@@ -112,7 +112,8 @@
 
         #define CODA_OSS_disable_warning_system_header_push \
             CODA_OSS_disable_warning_push \
-            CODA_OSS_disable_warning(-w)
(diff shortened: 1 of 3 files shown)


JDanielSmith (approved):
Personally, I don't like the macros to explicitly disable certain warnings (`CODA_OSS_FUNCTION_CALL_MISSING_ARG_LIST` and `CODA_OSS_DISABLE_UNREACHABLE_CODE`) as I'd rather see code changes made (if at all possible).  The generic push/pop are fine, although not too useful because the syntax to turn off warnings is different.

PR #525, opened 2021-09-09
Swig fixes

kdwilhelm:
The SWIG-generated Python bindings have gotten out of sync with the C++ code they wrap, and SWIG could no longer run successfully. Several issues were fixed in the SWIG *.i files, and a new Python module `config-python` was added to handle compatibility issues with `static_assert` and the `final` class specifier. Other SWIG *.i files should `%include "config.i"` and add `config-python` as a dependency to obtain these fixes.

Github Actions recently added SWIG support, and this was enabled for most of the CI builds to prevent future issues.


JDanielSmith (approved):
I've sometimes written `/*final*/` to remind myself that SWG doesn't like `final`.

Maybe we should add a `SWIG_FINAL` macro?  Or even just `FINAL`? (The "correct" `CODA_OSS_FINAL` is rather wordy.)

PR #572, opened 2022-05-12
WIP: Changes from CODA autofocus updates

konradrauscher:


modules/c++/mt/include/mt/GenerationThreadPool.h
```diff
@@ -105,6 +107,35 @@ namespace mt
 	    waitGroup();
 	}
 
+    
+    /*!
+     *  \brief Runs a given operation on a sequence of numbers in parallel
+     *
```
JDanielSmith, line 113: *personally* I'd prefer to see new code use C++11's task-based threading rather than adding to code in **mt**; but have at it ...

konradrauscher: Would that mean something like std::async? I definitely need more fine-grained control than that for the code I'm using this for.

konradrauscher: I definitely would rather use standard C++ threading utilities, but C++ doesn't have an official thread pool class so I had to use the CODA one.

PR #594, opened 2022-08-29
Added handler accessor to MinidomParser for inheriting modules

chvink:


chvink:
ugh dang swig is causing issues

chvink:
OK fixed, though things required me to regenerate all the swig bindings. The two files changed by hand include:

- modules/c++/xml.lite/include/xml/lite/MinidomParser.h
- modules/python/xml.lite/source/xml_lite.i

PR #607, opened 2022-10-12
Fix install order for cmake-modified xerces header

dave-starinshak:
Need to install the cmake-modified (autoconf) header after installing the xercesc source tree. The source tree contains a default header at the same path that will clobber the cmake-modified copy otherwise.


kdwilhelm:
Looks good. I don't think this would happen from a fresh clone, but if you build with waf or try an in-tree CMake build then your source tree will end up having a config.h

PR #784, opened 2025-02-10
Parallel file reads through io::FileInputStreamOS

Klamptoo:
Parallel file reads through io::FileInputStreamOS for high-performance reading of large contiguous file blocks


adam-beauchamp (approved):
LGTM

PR #785, opened 2025-02-27
Add build tree manifest file and check equality in pipeline

holmesv3:
Sanity check that anything messing with the cmake build code isn't changing the build tree. If it is, the `.github/build.manifest` file should first be updated.


holmesv3:
Closing to figure out the differences on my own branch before spamming MR commits.

PR #786, opened 2025-02-28
Cmake updates and auto test

holmesv3:


holmesv3:
Through some testing I found this doesn't work properly when using the `--clean-first` option with CMake. 

Inspiration was originally taken from the [AWS C++ SDK auto-test](https://github.com/aws/aws-sdk-cpp/blob/main/generated/tests/accessanalyzer-gen-tests/CMakeLists.txt#L34-L38) code, but may have been oversimplified in translation.

PR #790, opened 2025-04-21
Some Updates

holmesv3:
- Add fixed size polynomial json
- Fix workflow cmake build configuration
- CXX standard configurable from higher level project


CMakeLists.txt
holmesv3, line 1: Enable different C++ std

PR #791, opened 2025-08-03
Add explicit implementations of polynomial velocity and acceleration

holmesv3:
My initial goal here was to add `velocity` and `acceleration` functions to the `Fixed1D` polynomial which were similar to the `OneD` versions. After looking closer at the [current implementation](https://github.com/mdaus/coda-oss/blob/036b17d68e937ebf45189cbb7728e82e2877eef3/modules/c%2B%2B/math.poly/include/math/poly/OneD.hpp#L156-L161) I thought it would probably be more performant to directly evaluate the velocity and acceleration from the existing coefficients instead of creating a new polynomial (or two in the case of the acceleration) and then evaluating that. 

I refactored the two functions in `OneD` poly and added them to the `Fixed1D`, then made a simple micro-benchmark test which checked the new  values matched the old then called each version a bunch of times and reported the total execution time of the loop (compiled with `-DCMAKE_BUILD_TYPE=Release` and `volatile` keyword on the loop calculation to prevent the compiler from optimizing out the value). 

Results with `OneD<double>` seem quite good
```
N_Trial = 10000, Poly Order = 5
------------ Old Stuff ------------
     Velocity: 650876 nanoseconds
 Acceleration: 1239871 nanoseconds
------------ New Stuff ------------
     Velocity: 41146 nanoseconds (~16x faster)
 Acceleration: 49426 nanoseconds (~25x faster)
```
Results with `Fixed1D<5, double>` are less convincing
```
N_Trial = 10000, Poly Order = 5
------------ Old Stuff ------------
     Velocity: 2652 nanoseconds
 Acceleration: 5187 nanoseconds
------------ New Stuff ------------
     Velocity: 2628 nanoseconds
 Acceleration: 2619 nanoseconds
```
* Since the `Fixed1D` is backed by a `std::array` the compiler can do a lot more work for us. Still with the `acceleration` method we get about double the performance.

### DIFF
--- a/modules/c++/math.poly/include/math/poly/Fixed1D.h
+++ b/modules/c++/math.poly/include/math/poly/Fixed1D.h
@@ -187,6 +187,36 @@ public:
     }
 
     /*!
+     *  Evaluate the 1st derivative of our polynomial at 'at'
+     */
+    _T velocity(double at) const
+    {
+        _T rv{};
+        double atPower = 1;
+        for (size_t i = 1; i <= _Order; i++)
+        {
+            rv += static_cast<double>(i) * mCoef[i]*atPower;
+            atPower *= at;
+        }
+        return rv;
+    }
+
+    /*!
+     *  Evaluate the 2nd derivative of our polynomial at 'at'
+     */
+    _T acceleration(double at) const
+    {
+        _T rv{};
+        double atPower = 1;
+        for (size_t i = 2; i <= _Order; i++)
+        {
+            rv += static_cast<double>((i - 1) * i) * mCoef[i]*atPower;
+            atPower *= at;
+        }
+        return rv;
+    }
+
+    /*!
      *  Integrate between start and end
      *
      */
--- a/modules/c++/math.poly/include/math/poly/Fixed1D.h
+++ b/modules/c++/math.poly/include/math/poly/Fixed1D.h
@@ -137,8 +137,8 @@ public:
         }
     }
 
-    inline size_t order() const { return _Order; }
-    inline size_t size() const { return _Order + 1; }
+    constexpr inline size_t order() const { return _Order; }
+    constexpr inline size_t size() const { return _Order + 1; }
 
     /*!
      *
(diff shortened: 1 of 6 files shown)


modules/c++/math.poly/include/math/poly/Fixed1D.h
```diff
@@ -186,6 +186,36 @@ template <size_t _Order, typename _T=double> class Fixed1D
         return rv;
     }
 
+    /*!
+     *  Evaluate the 1st derivative of our polynomial at 'at'
+     */
+    _T velocity(double at) const
```
adam-beauchamp, line 192: Can you add a unit test for this guy? (odd/even len coeff array)

modules/c++/math.poly/include/math/poly/Fixed1D.h
```diff
@@ -186,6 +186,36 @@ template <size_t _Order, typename _T=double> class Fixed1D
...
+            rv += static_cast<double>(i) * mCoef[i]*atPower;
+            atPower *= at;
+        }
+        return rv;
+    }
+
+    /*!
+     *  Evaluate the 2nd derivative of our polynomial at 'at'
+     */
+    _T acceleration(double at) const
```
adam-beauchamp, line 207: ditto

modules/c++/math.poly/include/math/poly/OneD.hpp
```diff
@@ -155,17 +155,32 @@ OneD< math::linear::VectorN<3, double> >::derivative() const
 
 template<typename _T>
 _T
-OneD<_T>::velocity(double x) const
+OneD<_T>::velocity(double at) const
```
adam-beauchamp, line 158: ditto

modules/c++/math.poly/include/math/poly/OneD.hpp
```diff
@@ -155,17 +155,32 @@ OneD< math::linear::VectorN<3, double> >::derivative() const
...
+      ret += static_cast<double>(i) * mCoef[i]*atPwr;
+      atPwr *= at;
+   }
+   return ret;
 }
 
 template<typename _T>
 _T
-OneD<_T>::acceleration(double x) const
+OneD<_T>::acceleration(double at) const
```
adam-beauchamp, line 173: you guessed it...

adam-beauchamp (changes requested):
Can you retire the Win 2019 builds from the build suite as well as addressing the other comments?

PR #806, opened 2026-09-09
Add function to set http headers in curl client

austinjtravis:
Recently high side AWS enabled Amazon EC2 Instance Metadata Service Version 2 (IMDSv2) for all instances which requires us to set a HTTP header when we make a request to the instance metadata endpoint.  This enables a way for us to set that header as well as adding a unit test around the two new methods

### DIFF
--- a/modules/c++/net/unittests/test_curl_handle.cpp
+++ b/modules/c++/net/unittests/test_curl_handle.cpp
@@ -1,0 +1,122 @@
+/* =========================================================================
+ * This file is part of net-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
+ *
+ * net-c++ is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


modules/c++/net/unittests/test_curl_handle.cpp
```diff
@@ -0,0 +1,122 @@
...
+        "Authorization: Bearer token123",
+        "User-Agent: TestClient/1.0"
+    };
+    handle.setHttpHeaders(multipleHeaders);
+    
+    // Test that headers can be replaced - should not throw
+    std::vector<std::string> newHeaders{"X-Custom-Header: value"};
+    handle.setHttpHeaders(newHeaders);
+    
+    TEST_SUCCESS;
```
vink-arka: How hard would it be to add a simple getter to cram the contents of the header curl_slist into a vector to verify header values are being sent in? Aside from that I think the tests we have here are all we can do short of standing up a mock endpoint or something and verifying headers are actually getting sent, which seems kind of extra.

austinjtravis: Just pushed up a getter and updated the unit test.  let me know what you think

vink-arka:
This looks good to me but I don't have access to my older accounts that have merge permissions anymore

modules/c++/net/unittests/test_curl_handle.cpp
```diff
@@ -0,0 +1,176 @@
+/* =========================================================================
+ * This file is part of net-c++
+ * =========================================================================
+ *
+ * (C) Copyright 2004 - 2017, MDA Information Systems LLC
```
holmesv3, line 5: @vink-arka @adam-beauchamp  does this copyright statement (and any others touched in this PR) need to be updated?

